PR #6, opened 2017-11-07
add example json with extra user input needed

jenshnielsen:


examples/qcodesrc_example.json
```diff
@@ -0,0 +1,15 @@
+{
+    "core":{
+        "loglevel": "DEBUG",
+        "default_fmt": "data/{date}/#{counter}_{name}_{time}"
```
Dominik-Vogel: Will this not create a folder ` C:\Users\Jens\src\data\data\...` i.e. with data twice?

jenshnielsen:
There is already an example anyway

PR #7, opened 2017-11-13
Decadac in Custom-Instr

ThorvaldLarsen:
Making the the decadac driver more streamlined so we only need to change things in the config file.

The code is not pretty with 'exec' but dont know a better way to code it - Just want to have all settings in the config file rather than both places.

Note new format in the instr.config file.

### DIFF
--- a/qdev_wrappers/T3/customised_instruments.py
+++ b/qdev_wrappers/T3/customised_instruments.py
@@ -284,7 +284,7 @@ class Decadac_T3(Decadac):
         '''
 
         for channelNum, settings in config.get('Decadac').items():
-            channel = self.channels(int(channelNum))
+            channel = self.channels[ int(channelNum) ]
             config_settings = settings.split(',')
 
             name = config_settings[0]
--- a/qdev_wrappers/T3/customised_instruments.py
+++ b/qdev_wrappers/T3/customised_instruments.py
@@ -246,8 +246,6 @@ class DacChannel_T3(DacChannel):
 
         fine_part = voltage - coarse_part
         fine_scaled = fine_part*200-10
-        print("trying to set to {}, by setting coarse {} and fine {} with total {}".format(voltage,
-              coarse_part, fine_scaled, coarse_part+fine_part))
         self.volt.set(coarse_part)
         slot.channels[fine_chan].volt.set(fine_scaled)
 
@@ -285,7 +283,7 @@ class Decadac_T3(Decadac):
         8: Fine or coarse mode channel
         '''
 
-        for channelNum, settings in config.get('Decadac'):
+        for channelNum, settings in config.get('Decadac').items():
             channel = self.channels(int(channelNum))
             config_settings = settings.split(',')
 
--- a/qdev_wrappers/T3/customised_instruments.py
+++ b/qdev_wrappers/T3/customised_instruments.py
@@ -244,7 +244,7 @@ class Decadac_T3(Decadac):
         # Couldnt get this to work with normal parameters so ended quite ugly using 'exec' to define input strings as methods
         config_file = config.get('Decadac')
 
-        for channnel, channelNum in enumerate(self.channels):
+        for channelNum, channnel  in enumerate(self.channels):
             config_settings = config_file[str(channelNum)].split(",")
 
             name = config_settings[0]
@@ -260,19 +260,20 @@ class Decadac_T3(Decadac):
             if  fine_mode == 'fine':
                 param = channel.fine_volt
             elif fine_mode == 'coarse':
-                param = channel.fine_volt
+                param = channel.volt
             else:
                 pass #  raise exception
 
+            channel.volt.set_step(step)
+            channel.volt.set_delay(delay)
+
             param.label = label
             param.unit = unit
-            param.set_step(step)
-            param.set_delay(delay)
             param.set_validator(vals.Numbers(rangemin, rangemax))
 
             if divisor != 1.:
                 # maybe we want a different label
-                setattr(self, name, VoltageDivider(param, divisor, label=label)
+                setattr(self, name, VoltageDivider(param, divisor, label=label))
                 param.division_value = divisor
                 param._meta_attrs.extend(["division_value"])
             else:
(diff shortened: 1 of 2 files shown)


qdev_wrappers/T3/customised_instruments.py
```diff
@@ -227,78 +227,66 @@ def __init__(self, name, address, config, **kwargs):
...
+            self.channels[chan].volt.set_step(step)
+            self.channels[chan].volt.set_delay(delay)
+            
+            
+            rangemin = float(config_settings[6])
+            rangemax = float(config_settings[7])
             vldtr = vals.Numbers(rangemin, rangemax)
             self.channels[chan].volt.set_validator(vldtr)
+            if config_settings[8] == 'fine':
+                if np.mod(chan,4)<3:
```
Dominik-Vogel: Why do you exclude the last channel of every slot?

ThorvaldLarsen: that's a mistake. Should have been less than or equal to 3 (or <4). Haven't noticed as we have never used the last channel.

Dominik-Vogel: But isn't mod(x,3)∈{0,1,2,3}? So that the if statement is always true?

ThorvaldLarsen: I take back what I said before. I think my idea was to not define the last two channels when in 'fine' mode but then made a wrong implementation. Also this isn't really needed as fine_volt already throws an error for the last two channels of each slot.
Maybe just remove the if statement for both 'coarse' and 'fine' mode? It definitely seems unnecessary.

Dominik-Vogel: ok

qdev_wrappers/T3/customised_instruments.py
```diff
@@ -227,78 +227,66 @@ def __init__(self, name, address, config, **kwargs):
...
-                raise ValueError(
-                    "Expected: min max. Got {}".format(chan_range))
-            else:
-                rangemin = float(range_minmax[0])
-                rangemax = float(range_minmax[1])
+            config_settings = config_file[str(chan)].split(",")
+            
+            step = float(config_settings[4])
+            delay = float(config_settings[5])
+            self.channels[chan].volt.set_step(step)
```
Dominik-Vogel: Why do you only set the step for the volt and not for the fine_volt?

ThorvaldLarsen: my thinking was that fine_volt refers back to volt. So by setting volt on all channels I will also define step on the fine_volt implicitly.

qdev_wrappers/T3/customised_instruments.py
```diff
@@ -243,49 +243,40 @@ def __init__(self, name, address, config, **kwargs):
...
-                        exec('self.{} = self.channels[chan].volt'.format(config_settings[0]))
-                        exec('self.{}.label = \'{}\''.format(config_settings[0],config_settings[1]))
-                        exec('self.{}.unit = \'{}\''.format(config_settings[0],config_settings[2]))
-
+            fine_mode = config_settings[8]
+
+            if  fine_mode == 'fine':
+                param = channel.fine_volt
+            elif fine_mode == 'coarse':
+                param = channel.fine_volt
```
ThorvaldLarsen: think this should be channel.volt for coarse mode.

qdev_wrappers/T3/customised_instruments.py
```diff
@@ -243,49 +243,40 @@ def __init__(self, name, address, config, **kwargs):
...
+            else:
+                pass #  raise exception
+
+            param.label = label
+            param.unit = unit
+            param.set_step(step)
+            param.set_delay(delay)
```
ThorvaldLarsen: set_step and set_delay should refer to channel.volt both in fine and coarse mode.

qdev_wrappers/T3/customised_instruments.py
Dominik-Vogel: well spotted.

qdev_wrappers/T3/customised_instruments.py
Dominik-Vogel: done

Dominik-Vogel:
Did you change any security settings @ThorvaldLarsen? I cannot push the changes....

ThorvaldLarsen:
Haven't changed anything... It is saying that maintainers are allowed to commit changes.

qdev_wrappers/T3/customised_instruments.py
```diff
@@ -322,8 +241,7 @@ def _set_fine_voltage(self, voltage):
             fine_chan = 3
         else:
             raise RuntimeError("Fine mode only works for Chan 0 and 1")
-        coarse_part = self._dac_code_to_v(
-            self._dac_v_to_code(round(voltage,2)-0.01) )
+        coarse_part = self._dac_code_to_v(self._dac_v_to_code(voltage-0.001))
```
Dominik-Vogel: @ThorvaldLarsen I think this should not be changed... I don't know how this got in here. But as I remember it, does the originalversion with the rounding come from T3. Can you confirm this?

qdev_wrappers/T3/customised_instruments.py
```diff
@@ -332,31 +250,78 @@ def _set_fine_voltage(self, voltage):
         self.volt.set(coarse_part)
         slot.channels[fine_chan].volt.set(fine_scaled)
 
-    def set_all(self, voltage_value, set_dcbias=False):
```
Dominik-Vogel: @ThorvaldLarsen  do we want to remove this?

ThorvaldLarsen: It is some old stuff that we never used. Lets remove it for now and get a clean driver to start from. Then maybe later there might be a few functions we want to add again.

qdev_wrappers/T3/customised_instruments.py
ThorvaldLarsen: I think this change is good. It avoids large voltage jumps in between setting the coarse and fine parts of the voltage. Not sure how it got here but reading it I like the change.

PR #8, opened 2017-11-13
Show num

ThorvaldLarsen:
New functionality to show_num:
- Allows subtracting an average from each row or from each collumn
- Can save matplotlib png files with such an average subtracted

show_meta:
New function that prints metadata to command line from a specified dataset and instruments. It is not a pretty function as parameters are not stored in the same place for all instruments. 
The function allows defining key_word to only print metadata that has key_word in label (allows printing a specific metadata value).

Also slightly off topic here:
Lowering resolution of png's as evernote couldnt handle file size.

### DIFF
--- a/qdev_wrappers/show_num.py
+++ b/qdev_wrappers/show_num.py
@@ -71,7 +71,7 @@ def show_num(id, useQT=False, ave_col=False, ave_row=False, do_plots=True, savep
             plots.append(plot)
     else:
         plots = None
-    return plots, data
+    return data, plots
 
 
 def show_meta(id,instruments,key_word=''):
--- a/qdev_wrappers/show_num.py
+++ b/qdev_wrappers/show_num.py
@@ -3,8 +3,13 @@ from qdev_wrappers.file_setup import CURRENT_EXPERIMENT
 from qcodes.plots.pyqtgraph import QtPlot
 from qcodes.plots.qcmatplotlib import MatPlot
 
+def check_experiment_is_initialized():
+    if not getattr(CURRENT_EXPERIMENT, "init", True): 
+        raise RuntimeError("Experiment not initalized. "
+                           "use qc.Init(mainfolder, samplename)")
 
-def show_num(id, useQT=False, ave_col=False, ave_row=False, do_plot=True, savepng=False):
+
+def show_num(id, useQT=False, ave_col=False, ave_row=False, do_plots=True, savepng=False, **kwargs):
     """
     Show  and return plot and data for id in current instrument.
     Args:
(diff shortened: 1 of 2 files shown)


ThorvaldLarsen:
Actually we would really like to have one more feature for show_num: Modify the colorscale range of 2D plots before saving a png.

qdev_wrappers/show_num.py
```diff
@@ -4,46 +4,110 @@
 from qcodes.plots.qcmatplotlib import MatPlot
 
 
-def show_num(id, useQT=False, do_plots=True, **kwargs):
+def show_num(id, useQT=False, ave_col=False, ave_row=False, do_plot=True, savepng=False):
```
Dominik-Vogel: Did you change `do_plots` to `do_plot` on purose or is this a typo? We might consider keeping the old version to not break compatibility for everybody else.

qdev_wrappers/show_num.py
```diff
@@ -4,46 +4,110 @@
...
         id(number): id of instrument
-        do_plots: Default False: if false no plots are produced.
-        **kwargs: Are passed to plot function
+        useQT (boolean): If true plots with QTplot instead of matplotlib
+        ave_col (boolean): If true subtracts average from each collumn
+        ave_row (boolean): If true subtracts average from each row
+        savepng (boolean): If true saves matplotlib figure as png
 
     Returns:
-        data, plots : returns the plot and the dataset
```
Dominik-Vogel: Same for this change: maybe it makes it a little more convenient to have this change, but we need to consider if it is worth breaking compatibility here. Let's talk later so that I can learn how everybody is using show_num...

qdev_wrappers/show_num.py
```diff
@@ -4,46 +4,110 @@
...
+        plot, data : returns the plot and the dataset
+
     """
-    if not getattr(CURRENT_EXPERIMENT, "init", True):
+    if not CURRENT_EXPERIMENT["init"]:
```
Dominik-Vogel: I do not understand why we should consider the experiment as initialized, when `CURRENT_EXPERIMENT` does not have an entry for `init`

qdev_wrappers/show_num.py
```diff
@@ -4,46 +4,110 @@
...
+
+def show_meta(id,instruments,key_word=''):
+    '''
+    Print meta data in command line for datafile id.
+    id (int): 
+    instruments (list): List of strings with instruments names for which metadata is printed.
+    key_word (string): Optional - String in label of metadata if printed.
+    '''
+
+    if not CURRENT_EXPERIMENT["init"]:
```
Dominik-Vogel: I'll refactor it out...

qdev_wrappers/show_num.py
ThorvaldLarsen: I think the version of show_num I had was slightly different than the one in qdev-wrappers. I have no preference for naming so feel free to change.

qdev_wrappers/show_num.py
ThorvaldLarsen: again no preference for me. Feel free to change this.

qdev_wrappers/show_num.py
ThorvaldLarsen: This part was in show_num before i started modifying it. Not entirely sure what its purpose is.

Dominik-Vogel:
On the little longer run, we could consider cleaning the structure of show_num up by separating the editing features, like averaging and the pure display features. But I think for now this looks fine.

qdev_wrappers/show_num.py
Dominik-Vogel: using `getattr` you can supply a default value (here `True`) if the key is not present in the dictionary. So I guess it is an improvement we should keep.

ThorvaldLarsen:
I agree that this is probably not the best and there are other feeatures like these that we would like. But this is extremely useful functions for us hence the quick implementation. Hopefully this will all be part of the GUI that is being made.

Dominik-Vogel:
If you have another look and see that it does not break anything we can merge this PR..

ThorvaldLarsen:
we are only using these functions in command line so there is nothing that can break by changing names or output order. Looks good to me!

PR #13, opened 2017-11-22
Adding Aliases to the QDac

Dominik-Vogel:
Adding aliases to the QDac in analogy to the implementation of the Decadac.


ThorvaldLarsen:
Looks good! 
Only thing is that there is missing a part for the Qdac in the instr.config file. I think we want to add it now rather than having to commit it later.
I guess it is gonna be pretty long with 48 lines... The other option is to make the driver look and see if there is a line for a specific channel before setting up the new parameter. Think we can just do whatever is easiest for now.

Dominik-Vogel:
You are totally right. We should in both cases only add the aliases for specified channels and not require the user to specify all of them.

PR #14, opened 2017-11-23
Rename T3 to cQED

Dominik-Vogel:


Dominik-Vogel:
@ThorvaldLarsen as I understand `step` and `delay` are properties of any parameter and are used by the `sweep` method, so they should also be used for the QDAC.

PR #18, opened 2017-11-23
add natalie scripts to the wrappers in the examples folder

nataliejpg:
This pr adds the scripts I use with the wrappers to the repository as examples and adds the init file as a template.  In the process I did some refactoring of the examples folder although the work for making it easy to use and intuitive is by no means complete. 

The main outstanding questions are
- where should the experiment run (ie what should the path be in spyder when you start it as this is where the qcodes jspn configs should live)
- where should out instrument and calib dict config templates (.config) live? I think we have concluded that when the init script is run it should probably make a copy in the local file and that should be used from then on (and of not found then the default/template one used). 
- where should the template init script live? I imagine we should make an amalgamation of the one I have added and the one in cQED which shouldn't be too hard but then where to put it when it's done?
- where should customised instruments live?

Most of these questions porbably have the same answer but I don't know what that answer is so in the interim I have out these things in subfolders of the exaples folder and I guess next time I measure I will choose a way and go with that but I imagine that in the meantime @ThorvaldLarsen will have chosen a way (I imagine by using cQED folder) and be going with that and it would be fab if we could converge. @Dominik-Vogel, keep us from fighting won't you?


nataliejpg:
this is replaced by #21 and decision that more customised scripts should live locally or on personal repositories.

PR #21, opened 2017-11-29
Make template folder

nataliejpg:
Changes cQED into a templates folder which one would then copy to a local drive of the computer when installing the wrappers and edit. I'll test it out as soon as I'm measuring so maybe best to wait to merge until then.


qdev_wrappers/templates/initialisation_script.py
```diff
@@ -4,34 +4,46 @@
...
+# import locally customised instruments from local_instruments
+from local_instruments import my_very_local_VNA
+
+# import wrappers
+from qdev_wrappers.file_setup import CURRENT_EXPERIMENT, my_init, close_station
+from qdev_wrappers.configreader import Config
+from qdev_wrappers.show_num import show_num, show_meta
+from qdev_wrappers.sweep_functions import do1d, do2d, do0d
+from qdev_wrappers.transmon import *
+from chickpea import Segment, Waveform, Element, Sequence
```
ThorvaldLarsen: Should this be part of the template? I dont think people by default installs chickpea.

nataliejpg: sure, I can remove it.

ThorvaldLarsen:
This looks good to me.

One thing we can decide down the line (or maybe just up to the individual user): It seems to me that having local drivers call the custom instruments drivers which then calls the Qcodes driver gets a bit too convoluted. Maybe better to just call the Qcodes driver directly and then copy the parts of the custom instr driver that you need.

nataliejpg:
@ThorvaldLarsen yh fair enough. I would leave it up to the user but I'm happy to change the example. I'll do that and then wait for @Dominik-Vogel  to comment before merging

ThorvaldLarsen:
Sounds good.

ThorvaldLarsen:
Two more suggestions: 
- Can we change the alazar controller import in the experiment init to import from the qdev wrappers controllers instead of the Qcodes?
- Possibly make the Local_instruments file into a folder called Local_Drivers instead with a GS200_local.py file in it? Following our discussion on the custominstruments folder format.

nataliejpg:
I'll change the alazar controller import but how about we merge this pr after that and then make another for refactoring customised instruments into individual files and in the same pr turn local_instruments into local_drivers? I think the turnaround should be pretty quick and we could have it done by the end of Monday. Just feels to me like those tasks fit together as one a bit better and I'd like to get this merged.

nataliejpg:
@Dominik-Vogel any thoughts? Tbh I would love to get this merged Monday and it would have the benefit of meaning you could use it on Tuesday when setting up the new measurement computer...

ThorvaldLarsen:
That sounds good. I can merge this Monday morning if it's not done before then.

nataliejpg:
or tbh I could merge now and then I can easily refactor the customised instruments this eve so that by monday it's all done. Obviously it won't be tested and although it seems likely I'll be able to test on monday it's not 100%?

ThorvaldLarsen:
That's fine with me. If you want to test anything monday it's fast on the new T3 computer which is not used next week.

PR #22, opened 2017-11-30
Organise examples

nataliejpg:
Reorganise examples folder, mainly deleting examples which don't actually use the wrappers and somewhat tidying up those that do so that they are all python scripts which can be run with spyder.

Currently there:
-  examples for how to do pulse building given the use of the calibration config and the general file setup etc of the wrappers done using an init function. It relies on having chickpea installed. Examples include using functions to build variants of the standard rabi and ramsey (etc) sequences and also gated sequences like ALLXY
- examples for measuring using the do1d, do2d etc wrappers
- examples of deca measuring using the customised parameter names and majorana conductance wrapper

Would be nice to add:
- an extension to the vna example to flesh out how spectroscopy mode works with all the latest bells and whistles
- an example of using some of the analysis and plotting functions when they have been cleaned up a bit
- some more examples of how the wrappers facilitate multi qubit measurement eg uploading to the awg, setting things on alazar and microwave source concepts.

@Dominik-Vogel @ThorvaldLarsen  thoughts?

ps. this branch is based off make_templates_folder so it probably makes sense to hold off merging until that is merged in #21 or to rebase it


Dominik-Vogel:
This is really nice @nataliejpg! We should consider converting the open points to an issue so we can get this PR merged

nataliejpg:
I'll make an issue and then we can merge what we have? On most of the 'things to add' the actually code will probably change quite a bit soon so it's maybe not worth writing the examples just yet.

nataliejpg:
the only other thing is if the file structure makes sense and is useable. I'd vote for merging and then we can change it if we realise it needs it

nataliejpg:
issue created #29

ThorvaldLarsen:
File structure makes sense to me.

Issue #23, opened 2017-11-30
Reformatting custom instruments


ThorvaldLarsen:
I think we should reformat the custom instruments file into a folder instead i.e. a Custom instruments folder that has files such as Decadac_cQED, ZNB_cQED, qDac_cQED... This way it is easy to get an overview of what instruments have extended and in PR's it is easy to see which instruments are affected.

PR #24, opened 2017-12-03
split customised_instruments into files

nataliejpg:
Solves issue  #23 

Needs a few decisions and ideally also a check done on real instruments (although I haven't changed any actual code so it should be fine)

- [x] some files have more than one class in (eg alazar has both the alazar and the acq controller and the Decadac has channel, slot and decadac in) should these be split more or is it nicer to group them? If so then can someone double check I've grouped decadac correctly (and not missed that something should go with the qdac)?
- [ ] naming: at the moment we have both the suffix 'cQED' and 'T10'. Should we abolish both and replace with 'customised' or is that just asking for trouble when people make conflicting drivers in the future and more than one is needed...?
- [x] QDac driver looks to me like it wouldn't actually work, has it been tested?
- [x] We should get rid of the ManualParamter as it is depreciated

@ThorvaldLarsen @Dominik-Vogel thoughts?


Dominik-Vogel:
Great @nataliejpg ,
- I think grouping makes perfect sense. (Same as in the original base instruments)
- I think calling them _customized is a good idea: The case of having conflicting modifications is relatively rare as compared to the advantage of everybody using the same updated instrument drivers/features. So that means we are striving for a unified set of customized instruments. If there is a customization that is conflicting, we can still iterate one level higher and in that case create a `InstrumentVerySpecialQT3Use`
- Are you insulting my dear qdac driver? ;-). You are right, there have been some updates to it, that are questionable, but it passed the basic tests at T7 and T11. @ThorvaldLarsen have you had any trouble with it? @nataliejpg can you be more specific, what parts you consider possibly disfunct?
- Let's make an issue for the ManualParameter and get this PR merged

ThorvaldLarsen:
With the current file structure for data folders one can also get the id through the dataset location:
id  = int(data.location.split('\')[-1])
where data is a dataset loaded by qc.load_data()

nataliejpg:
@Dominik-Vogel turns out my linting is just wrong and the QDac driver is fine, my bad. I'll make an issua for manual parameter and then I just need to change cQED to _customised and then it's good to merge. I guess realistically the next time someone pulls it will be one of us so if any of the imports are wrong we can pretty quickly fix it.

nataliejpg:
#31

ThorvaldLarsen:
Can we come up with a shorter name extension instead of _customised? I think I will get annoyed by te long name (likely only me).
what about _ext for extension? as most of the drivers are actually extension building on top of the codes driver.

nataliejpg:
@ThorvaldLarsen I have zero preference. @Dominik-Vogel? Just let me know before I change it.

Dominik-Vogel:
I personally have no preference either. _ext is fine to me

ThorvaldLarsen:
Any chance we can merge this PR soon? Would be great to have it for update on T11.
I can take a look at the naming if you are in the middle of other stuff @nataliejpg.

nataliejpg:
I can do it now

nataliejpg:
maybe have a check @ThorvaldLarsen but I think it's good to merge and test now (or in the other order)

jenshnielsen:
Are you sure no one else are using these instruments who's code you would break? Perhaps providing aliases at the old location would be a good idea

nataliejpg:
@jenshnielsen Maybe we should work out who is actually using the wrappers at the moment (and who actually know how to pull updates) as while obviously we don't want to break things for anyone it is also quite a drag to make extra files which will hang around if there aren't people who actually need them. I feel like this sort of stuff will settle down pretty soon when the structure is a bit more set. The reason I am sceptical about the need for aliases is that customised_instruments only very very recently moved out of the very temporary cQED folder so this will only break things for people who have exactly started using these wrappers in the last week since then and have used the template init script (which is also very new)...

nataliejpg:
@jenshnielsen the short answer would be that I would be willing to bet that there isn't.

ThorvaldLarsen:
The only one possibly using it would be Fabio but I think he is using the one in SpinQubits folder. Will check with him and then merge after looking through the changes.

ThorvaldLarsen:
Just checked with Fabio. They are using local versions of the files so should be fine to merge.
If no last minute objection I will merge this PR.

PR #26, opened 2017-12-03
add (back) in data_num concept

nataliejpg:
This is a short term solution to the fact that datasets do not have an easily accessible id so you have to instead use the qc.data.data_set.DataSet.location_provider.counter to find out what the most recent data_set taken was. This number is then set as an attribute of the dataset so that if I want to use the dataset to create an analysis plot or similar it is easy to get the counter at the time this dataset was taken and use that as a title or when saving.

### DIFF
--- a/qdev_wrappers/sweep_functions.py
+++ b/qdev_wrappers/sweep_functions.py
@@ -85,7 +85,8 @@ def _do_measurement_single(measurement: Measure, meas_params: tuple,
             plot.save()
             if 'pdf_subfolder' in CURRENT_EXPERIMENT:
                 plt.ioff()
-                pdfplot, num_subplots = _plot_setup(data, meas_params, useQT=False)
+                pdfplot, num_subplots = _plot_setup(
+                    data, meas_params, useQT=False)
                 # pad a bit more to prevent overlap between
                 # suptitle and title
                 pdfplot.rescale_axis()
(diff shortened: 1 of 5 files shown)


Dominik-Vogel:
Please do not make any commits with "pep8 and ....". These are a nightmare to review, because the you can't see the changes in a diff. Best would be to split them up.

Dominik-Vogel:
I really feel with you @nataliejpg : it is extremely painful to find the counter id of a dataset. And it should not be. This should be changed in main qcodes, but even that is not straight forward. I'll try to find a solution to that.
I am not 100% happy with the solution in this PR either:
 - you are attaching an attribute to the DataSet object from outside of the methods, which is not a good practice
 - if you are doing this you should at least avoid the repetitions and do it in _do_measurement.
 - data_num is not an ideal name for the new attribute... (This could be anything)

jenshnielsen:
For now I would suggest adding it to the metadata dict which is already there in the dataset

ThorvaldLarsen:
With the current file structure for data folders one can also get the id through the dataset location:
id = int(data.location.split('')[-1])
where data is a dataset loaded by qc.load_data()

Dominik-Vogel:
What about just using a function in the wrappers that extracts the number from the filename, as it is already done for the get_latest? I think this would be the easiest solution for now, until the new dataset comes.

nataliejpg:
I'm really not attached to this method of doing it I just want it to be something easy and quick because it feels to me so much like it shouldn't even be an issue and it should be trivial to get the 'id' of a dataset which is what the counter effectively acts as. I can make a function which does it instead, I'll just have to take long deep breaths while I'm doing it.

nataliejpg:
and yes about pep8 @Dominik-Vogel, bad habit from being the only person working on a repository for so long :P

nataliejpg:
replaced by #45

PR #27, opened 2017-12-05
Qdev fitter

ThorvaldLarsen:
First draft for function to handle easy fitting of qcodes data files. This is not a finished version but wanted to make a PR to discuss the structure of the function - most likely there is a better way of doing this.

- This function has been tested and works for fitting 1D datasets doing T1 and T2 fits.
- Not tested: Samplefolder input for accessing files not in current experiment.
- Custom function input.


qdev_wrappers/qdev_fitter.py
```diff
@@ -0,0 +1,141 @@
...
+        # Function only made for a single dataname. Check there isnt more datanames in the dataset.
+        key_data = [key for key in data.arrays.keys() if "_set" not in key[-4:]]
+        if len(key_data) != 1:
+            raise RuntimeError('Dataset has more than one dataname. Choose dataname to be fitted from: {}.'.format(', '.join(key_data)))
+    else:
+        # Check input dataname is in dataset.
+        if dataname not in [key for key in data.arrays.keys() if "_set" not in key[-4:]]:
+            raise RuntimeError('Dataname not in dataset. Input dataname was: \'{}\' while dataname(s) in dataset are: {}.'.format(dataname,', '.join(key_data)))
+        key_data = [dataname]
+
```
Dominik-Vogel: Here I would check if fitfunction is a `callable`, so you can also pass a class to it for more complicated fit routines. 
Furthermore I am not sure whether it makes a lot of sense to have a string describing the function. What do you have in mind? Is it not that the only difference is the quotation marks?

qdev_wrappers/qdev_fitter.py
```diff
@@ -0,0 +1,141 @@
...
+    id (num): Id of dataset
+    fitfunction_in (str, function): Function data will be fitted to. Can be string for functions defined here or a function passed direclty defined elsewhere.
+    do_plots (bool): If plots are done. Will be saved in the analysis folder for the samplepath
+    dataname (str): String with dataname in the qcodes dataset
+    samplefolder (path): Path to samplefolder if not opening active measurements
+    p0 (tuple): Initial parameters for fit
+    kwargs: Passed to scipy.optimize.curve_fit when fitting.
+    '''
+
+    # Load data from active samplefolder or input samplefolder.
```
Dominik-Vogel: Ok, I'll definitely refactor this out, as the next lines get used in three different places now...

qdev_wrappers/qdev_fitter.py
```diff
@@ -0,0 +1,141 @@
...
+    else:
+        print('Invalid input. Only takes strings of predefined fucntions or function input.')
+
+
+    # Do fits separately for 1D and 2D datasets
+    keys_set = [key for key in data.arrays.keys() if "_set" in key[-4:]]
+
+    # Do fit and plot for 1D data
+    if len(keys_set) == 1:
+        xdata = getattr(getattr(data, keys_set[0]), 'ndarray')
```
Dominik-Vogel: I think you should be able to replace the outer getattr by a simple dot

qdev_wrappers/qdev_fitter.py
```diff
@@ -0,0 +1,141 @@
...
+            return popt, pcov
+
+    # Not implemented
+    if len(keys_set) == 2:
+        if xname==None:
+            raise RuntimeError('Need fitting dimension for 2D datasets. Choose fit dimension from: {}.'.format(', '.join(keys_set)))
+        raise RuntimeError('Fitting not ready for 2D datasets.')
+
+
+def check_experiment_is_initialized():
```
Dominik-Vogel: This function is confusing. How is it different from 'check_experiment_is_initialized()' in show_num.py? If it is not we should rather import it, or move it to a different place and import it.

Dominik-Vogel:
Great that you are starting to fit with qcodes @ThorvaldLarsen! Besides my boring comments above, we should consider the following:
 - beware that QCoDeS generally does not include any fitting routines.
 - make sure that you keep the connection to QCoDeS as loose as possible. This will make your future life a lot easier, because to maintain a code base becomes (if not designed well) exponentially more difficult with growing size and fitting will definitely grow in the futuer. In my opinion, one can make a clear cut between data acquisition and fitting with a minimal interface. It is a little bit unfortunate that qcodes does not perfectly separate fitting and plotting, so that it also becomes a little more difficult to separate the fitting out. You should ask yourself: On what qcodes objects does my code actually depend? What would it look like, if you wrote a fictive pure fitting framework agnostic of qcodes and then make a thin layer that connects qcodes up to your fitting framework.

ThorvaldLarsen:
I basically just want a function that I can pass a measurement number to (like show_num) and then get a fit back and a nice plot with data at fit shown. 
To do this I use Qcodes for two things: Data load and qc.Matplot. The loading is an easy way to pass and handle the data. Fitting is taking advantage of the nice features of qc.Matplot fixing labels and units and qc.Matplot.rescale() to change units (0.001V -> 1mV).

I agree that it would be nice to separate more but that would also mean that I would have to rewrite these functions which to me seems like a bigger project.

qdev_wrappers/qdev_fitter.py
ThorvaldLarsen: I want an easy way to acces functions defined within the file (T1,T2) instead of having to pass a function.
Most often we are using a same few functions over and over. However, I also want to be able to pass a function for that one odd time where we need to fit something different.

qdev_wrappers/qdev_fitter.py
ThorvaldLarsen: This is the exact same code as in show_num and show_meta. Refactoring to somewhere else is fine for me. Not sure where it should go though. Maybe just in the show_num file?

qdev_wrappers/qdev_fitter.py
ThorvaldLarsen: its the same one. Just a copy as i didnt think to import it.

qdev_wrappers/qdev_fitter.py
Dominik-Vogel: I am not sure what exactly you mean. You can do exactly what you described if you define the function in the beginning of the file:
```
def fitfun(x,p1):
   return sin(p1*x)

def dothefit(fitfun, parmeters, constraints):
   ...
```

ThorvaldLarsen: I have fitfun defined as in your example. Now I want to run dothefit() from outside the file where fitfun is not known (I dont want to import all functions in the file).

Dominik-Vogel: ok, now I get, what you meant was inconvenient and why you chose to circumvent it like this. I think there are two options you might find interesting to consider:
 - you can wrap everything in a class and define your fit function as a static method
 - you can make a module, so that you, just you would do with numpy, import the whole module and then use the dot notation. eg 
```
import ....thorvaldsFit as fit
from fit import dothefit
...
dothefit(fit.fitfun,parameters, constraints,...)
```
- a third option is to do a star import, which, as you know is not considered best practice, but which is, to my mind, a better solution than the workaround with the string for the name of the function.

ThorvaldLarsen: I dont like the start import but the other two options are interesting. Will take a look at them.

ThorvaldLarsen:
Redid the function using classes instead. Dropped the idea of allowing function input and instead defining a class for the fit function. This allows to have some metadata about the function (name and parameter names) for plotting.

Plotting works very nicely with the current version but the function is pretty unreadable... It is based around getting the units for a parameter from the axis label and getting the scaling (micro/mili/kilo...) from the axis major_formatter function which is modified by rescale_axis() in Qcodes matplot.

ThorvaldLarsen:
I know that this is not the prettiest function but as it works really well in its current state i would suggest merging it now.
No other function is calling from it so it can't break anything.

PR #30, opened 2017-12-06
show_num and show_meta without experiment init

ThorvaldLarsen:
Allow opening files from input path to samplefolder without init script. 
I have tested this.


Dominik-Vogel:
This looks like a good improvement to me. Maybe we could refactor data loading out into a function.

ThorvaldLarsen:
I agree that the data loading part should probably be refactored out to a different function. However, is it okay to merge this for now and then we can make a new PR with the refactored version?

nataliejpg:
looks good to me in terms of functionality but there is quite a lot of identical code in both functions. Is this what you mean @Dominik-Vogel by refactoring data loading out? if so then I would agree but also vote to merge this now and make an issue out of it so that it gets solved when someone has time

Dominik-Vogel:
Sounds good to make an issue and merge it now. The refactoring functions out should be a quick and painless thing to do....

PR #32, opened 2017-12-06
Custom VNA

Dominik-Vogel:
This PR requires PR887 from qcodes (https://github.com/QCoDeS/Qcodes/pull/887).

In this PR the features of the custom VNA code shall be gathered.

- [x] adding `FrequencySweepMagSetCav`
   a trace that after every readout sets the readout frequency around the peak of the spectrum.
- [x] auto activation of the measured trace and resetting to its original status (main qcodes pr)
- [x] addtional features from custom vna. What is there around?


nataliejpg:
could we confirm the status of this @ThorvaldLarsen? @Dominik-Vogel? Just wondering who is working in it at the moment.

Dominik-Vogel:
The code is ready to test

nataliejpg:
@Dominik-Vogel just reading the code I'm wondering why so much of it appears new, to me it looks like exactly the version already in all_wrappers with the addition of FrequencySweepSetMagCav and the removal of singleS21? Am I missing something? I know it seems like maybe overkill but would it be possible to make a quick example (or add some more comments to make clear what FrequencySweepSetMagCav is?). Reading the code it appears to me to set the 'readout_freq' based on the minima of the return from a FrequencySweep. It isn't clear to me what this will do if the channel is not a spectroscopy channel (as it looks to me like every ZNB_channel will now have a trace_mag_SetCav param but not every channel has a readout_freq)... I just feel like an example would really really help.

Dominik-Vogel:
@nataliejpg that is correct. It is supposed to be the same, except for the new channel class and the new sweep parameter. The channel class and sweep parameter needed changes of the qcodes driver, so this pr only works with PR887.
The singleS21 was lost in the chaos of thousands of versions of this driver (I picked a different one from the one you decided to use). Good that you noticed. I will add it.
What functionality are you missing? Or why are you surprised?

qdev_wrappers/customised_instruments/ZNB_ext.py
```diff
@@ -0,0 +1,145 @@
...
+                'readout_power',
+                label='{} Readout power'.format(self.vna_parameter()),
+                unit='dBm',
+                set_cmd=partial(self._set_readout_pow, self._instrument_channel),
+                get_cmd=partial(self._get_readout_pow, self._instrument_channel),
+                get_parser=int,
+                vals=vals.Numbers(-150, 25))
+
+    def _set_readout_freq(self, channel, freq):
+        self.write('SOUR{}:FREQ:CONV:ARB:EFR1 ON, 0, 1, {:.6f}, '
```
ThorvaldLarsen: I think this should also be {:.7f}.

ThorvaldLarsen:
Can we add default values for readout_freq=6GHz and readout_pow=-60dBm that is set when the spectroscopy channel is added.? When adding the spectroscopy channel these needs to be set to add the channel correctly.

Also this is the last point from issue #48.

qdev_wrappers/customised_instruments/ZNB_ext.py
Dominik-Vogel: done

Dominik-Vogel:
ok standards are added

Dominik-Vogel:
as soon as this is tested we can merge the two PRs

qdev_wrappers/customised_instruments/ZNB_ext.py
```diff
@@ -0,0 +1,152 @@
...
+        self.write('SOUR{}:POW2:OFFS {:.3f}, ONLY'.format(channel, pow))
+
+    def _get_readout_pow(self, channel):
+        return self.ask('SOUR{}:POW:GEN1:OFFS?'.format(channel)).split(',')[0]
+
+    # this is a quick and dirty work around to get scaling of the plot
+    # in mod 3 powers
+    def _set_format(self, val):
+        super()._set_format(val)
+        if val == 'MLIN':
```
ThorvaldLarsen: if val == 'MLIN\n':

Dominik-Vogel: 👍

qdev_wrappers/customised_instruments/ZNB_ext.py
```diff
@@ -0,0 +1,152 @@
+import logging
+import time
+from functools import partial
+import numpy as np
+
+# QCoDeS imports
+from local_instruments.ZNB_new import ZNB, ZNBChannel
```
ThorvaldLarsen: Should call from qcodes instead of local instruments

qdev_wrappers/customised_instruments/ZNB_ext.py
```diff
@@ -0,0 +1,152 @@
...
+#from qcodes.instrument_drivers.rohde_schwarz.ZNB import ZNB, ZNBChannel
+from qcodes.utils import validators as vals
+
+log = logging.getLogger(__name__)
+
+class ZNBChannel_ext(ZNBChannel):
+    def __init__(self, parent, name, channel):
+        super().__init__(parent, name, channel)
+
+        if self. channel_name() == 'B2G1SAM':
```
ThorvaldLarsen: Remove spaces in self. channel_name()

ThorvaldLarsen:
@nataliejpg there are some changes to transmon/sequencing/basic.py in this PR. Is that a mistake?

nataliejpg:
@ThorvaldLarsen yes at the very least they shouldn't be in this pr. I can pull them out and put them as a local script or sth today. Also did the import channels from local part get changed to import from qcodes? you wrote a comment on it but it's still there (could well be my own github illiteracy).

Dominik-Vogel:
My bad, somehow my changes did not get pushed. But now its there

ThorvaldLarsen:
@Dominik-Vogel Can you discard the changes to the sequencing folder in this PR?

Then I will try to get this tested today/tomorrow with the new VNA parameter stuff also.

PR #37, opened 2017-12-11
 do0d png save and use_threads=False

ThorvaldLarsen:
- Added png save for do0d
 - changed default value of use_threads to False


qdev_wrappers/sweep_functions.py
```diff
@@ -83,18 +83,30 @@ def _do_measurement_single(measurement: Measure, meas_params: tuple,
...
+                    pdfplot.save("{}.pdf".format(title))
+
+                if 'png_subfolder' in CURRENT_EXPERIMENT:
+                    # Hack to save PNG also
+                    title_list_png = plot.get_default_title().split(sep)
+                    title_list_png.insert(-1, CURRENT_EXPERIMENT['png_subfolder'])
+                    title_png = sep.join(title_list_png)
+
+                    plt.savefig("{}.png".format(title_png),dpi=500)
+
```
jenshnielsen: I would do this as (untested), This should change below in doND too

```
if 'pdf_subfolder' in CURRENT_EXPERIMENT or 'png_subfolder' in CURRENT_EXPERIMENT:
    title_list = plot.get_default_title().split(sep)
    title_list.insert(-1, CURRENT_EXPERIMENT['pdf_subfolder'])
    title = sep.join(title_list)
    if 'pdf_subfolder' in CURRENT_EXPERIMENT
         pdfplot.save("{}.pdf".format(title))
   if 'png_subfolder' in CURRENT_EXPERIMENT:
        pdfplot.save("{}.pdf".format(title))
``` 

And then change the pdfplot variable to something more descriptive

jenshnielsen:
Any particular reason for changing the default of use_threads

Dominik-Vogel:
It would be great to refactor the changes into a function. As not being the author it is difficult to see, if there are any changes or if it is only copy paste.

ThorvaldLarsen:
We have a type of measurement where we call for two measurements from the VNA. These has to be done sequentially as it are measurements on the same instrument.

We do this one all the time and every once in a while I forget to add the use_threads=False to the do1d().

ThorvaldLarsen:
@Dominik-Vogel Yes this is a direct copy past from the do_measurement into _do_measurement_single. This was the fastest fix on T11 but not the prettiest. Refactoring would be good.

ThorvaldLarsen:
I just updated a new version with the marplot part refactored out. This was tested on T11.

I know this is still not super nice, but think it might be a bigger thing to get all-wrappers plotting code nice. Maybe we can merge this unless there are glaring mistakes and then come back to plotting at a later point?

PR #38, opened 2017-12-12
only specify named channels in config for qdac

### LINKED ISSUE #36: Qdac_ext not working
- want to change it so that it loops over entries in the config file instead of number of channels.
- voltage parameter is not channel.volt as for decadac.

Dominik-Vogel:
fixes #36

### DIFF
--- a/qdev_wrappers/customised_instruments/QDAC_ext.py
+++ b/qdev_wrappers/customised_instruments/QDAC_ext.py
@@ -13,10 +13,9 @@ class QDAC_ext(QDac):
         super().__init__(name, address, **kwargs)
 
         # same as in decadac but without fine mode
-        config_file = config.get('QDAC')
-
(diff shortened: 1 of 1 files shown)


ThorvaldLarsen:
We also need to fix how it calls the parameter. The qdac driver does not have channels.

ThorvaldLarsen:
just noticed that there might be a different version of the qdac driver with channels... maybe we just need to use that one instead. I dont know how that one works though.

ThorvaldLarsen:
@Dominik-Vogel I tested this at QT1 with Lucas. Only thing more that needs to be changed is line 29:
param = channel.volt -> param = channel.v

Dominik-Vogel:
@ThorvaldLarsen thank your for testing and pointing that out. This is fixed now and we can merge.

PR #43, opened 2017-12-16
Replace '−' with '-' in string to float conversion

ThorvaldLarsen:
Weird problem when converting string into float. On T2 the minus sign of the string is a sign not understood by float().


ThorvaldLarsen:
I am going to merge this one as it is only affecting the fitting tool that no one else is using yet.

PR #51, opened 2017-12-19
Config load

ThorvaldLarsen:
Change the name found when loading from the config file. This allows to have settings for 2 decadac's in the config file by setting [deca1] followed by channels settings and then [deca2] followed by settings for decadac 2

The template init script should also be changed so that when initialising the instrument the name given matches the config file. In the above case that would be:
deca1 = Decadac_ext('deca1', 'ASRL1::INSTR', instr_config,terminator='\n')
deca2 = Decadac_ext('deca2', 'ASRL2::INSTR', instr_config,terminator='\n')


ThorvaldLarsen:
Also fixed the decadac initialisation to have "terminator='\n'" as input. However, this should probably go into the main qcodes driver for the decadac as this is fixed by the decadac.

PR #52, opened 2017-12-20
Merge all wrappers into master

jenshnielsen:


ThorvaldLarsen:
The All-wrappers branch is working nicely on T2, T3, T11, QT1, and QT2.

I think the current version would be a good point to move the code to the master branch and then make sure the update goes well on all setups (hopefully just a simple pull on all computers).

PR #53, opened 2017-12-20
Fixing the QDac aliases

### LINKED ISSUE #50: QDac numbering
There is a bug with the numbering of channels on Qdac. Python starts counting channels at 0 while the channel outputs on the qdac are named 01-48.

So when setting in the config file 0 = name, .... The qdac output channel set by this line is Ch01.

Dominik-Vogel:
This fixes #50 the problem of the qdac channels being numbered starting at 1 while channels are numbered from zero.
In the `instr.config`
```
0 = Q1lplg,Q1 left plunger,V,1,0.1,0.1,-10,10
```
leads to the assignment of channel 1.
With this PR the config has be written as:
```
ch01 = Q1lplg,Q1 left plunger,V,1,0.1,0.1,-10,10
```
(note 0 changed to 1)
Here `ch` is added to throw an error if the config file is unchanged, instead of simply shifting all channels and possibly breaking devices.

### DIFF
--- a/qdev_wrappers/customised_instruments/QDAC_ext.py
+++ b/qdev_wrappers/customised_instruments/QDAC_ext.py
@@ -15,7 +15,13 @@ class QDAC_ext(QDac):
         # same as in decadac but without fine mode
 
         for attribute, settings in config.get('QDAC').items():
-            channel = getattr(self, attribute)
+            try:
+                channel = getattr(self, attribute)
+            except AttributeError as e:
+                raise AttributeError(e.message +
+                                     'config file implementation has changed:\n'+
+                                     '\'1 = ... -> chan02 = ...\' make sure to'+
+                                     ' increment the channel number!')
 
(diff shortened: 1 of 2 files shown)


ThorvaldLarsen:
Just tested this. We need to change the config file a bit:
[QDAC]
# this is an example on how to use this config file
ch01 = LQcut,LQ cutter,V,1,0.1,0.1,-10,10
ch02 = RQlplg,RQ left plunger,V,1,0.1,0.1,-10,10
ch03 = RQcut,RQ cutter,V,1,0.1,0.1,-10,10
ch04 = RQrplg,RQ right plunger,V,1,0.1,0.1,-10,10
ch05 = testchan,Test,V,1,0.1,0.1,-10,10

Otherwise everything works.

Dominik-Vogel:
ok great you discovered that.. I have changed it and will change the description also

PR #54, opened 2017-12-20
Subplots with mulitparameters need to be fixed

Dominik-Vogel:
Initial attempt

### DIFF
--- a/qdev_wrappers/plot_functions.py
+++ b/qdev_wrappers/plot_functions.py
@@ -77,15 +77,18 @@ def _plot_setup(data, inst_meas, useQT=True, startranges=None):
                     fulltitle = ""
                 plot.subplots[j + k].set_title(fulltitle)
 
-    for j, i in enumerate(inst_meas):
+    subplot_index = 0
+    for measurement in inst_meas:
         if getattr(i, "names", False):
             # deal with multidimensional parameter
             for k, name in enumerate(i.names):
-                _create_plot(plot, i, name, data, counter_two, j, k)
+                _create_plot(plot, i, name, data, counter_two, subplot_index, 0)
+                subplot_index += 1
                 counter_two += 1
         else:
             # simple_parameters
-            _create_plot(plot, i, i.name, data, counter_two, j, 0)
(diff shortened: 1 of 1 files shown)


qdev_wrappers/plot_functions.py
```diff
@@ -77,15 +77,18 @@ def _create_plot(plot, i, name, data, counter_two, j, k):
                     fulltitle = ""
                 plot.subplots[j + k].set_title(fulltitle)
 
-    for j, i in enumerate(inst_meas):
+    subplot_index = 0
+    for measurement in inst_meas:
```
ThorvaldLarsen: we change i -> measurement.
This is not done for all calls below.

Otherwise this change makes lifeplotting work when calling multiple multi array parameters.

nataliejpg:
@Dominik-Vogel could you write a sentence about what is broken and what you are trying to fix?

nataliejpg:
or ref the issue this solves

ThorvaldLarsen:
This problem arose when doing measurements involving 2 alazar controllers. The plot function does not count the number of subplots correctly when a single measurements has more than one dataset. Following measurements will plot in already existing subplots instead of creating new ones.

Dominik-Vogel:
Sorry @nataliejpg for not having been more concise here, I was about to leave and pushed to code up to share it with @ThorvaldLarsen. It seemed to us as if the code was inconsistent in the indexing with counter_two, j and k. 
In the plotting function you were refering to subplots j+k only as a sum (except for setting the title), but when cycling through the parameters j is the index for the parameter and k the index for the subparameter (from multiparameter). So that if you have for example two multiparameters with two parameters each you would get:
```
j     k    j+k
1     1    2
1     2    3
2     1    3
2     2    4
```
with the result that subplot 3 gets overwritten

nataliejpg:
The problem was that you were too concise not that you weren't concise enough ;) thanks for the explanation, seems good to me

ThorvaldLarsen:
Still need the changes I mentioned above before merging.

ThorvaldLarsen:
This PR is superseded by #55.

PR #55, opened 2018-01-15
Fix subplot counter

ThorvaldLarsen:
Bug fix for subplot counter for multiparameter plotting.

This supersedes #54.


nataliejpg:
it's not exactly an easy function to read but I assume you've tested?  ;) If so I would say go ahead and merge (and also thank youuuuu)!

ThorvaldLarsen:
Agreed. This is why I think a refactoring of some of these functions would be useful.

And yes, this has been tested.

nataliejpg:
well then I leave you the joy of pressing the big green button

PR #56, opened 2018-01-16
remove obsolete sweep_helpers.py

Dominik-Vogel:
Sweep helpers seems to be outdated and superseeded by sweep_functions.py.


ThorvaldLarsen:
Yes, I dont use the sweep helpers at all. However, I think maybe Natalie is still calling them from other functions.

Dominik-Vogel:
They get called in 
 * `alazar_automation.py` lines 237, 312 
 * `alazar_rs_helpers.py` line 335 
 *  `vna_helpers.py` line 41

nataliejpg:
Yes they still get called in some of the automated functions I use. I also personally prefer the 'start', 'stop', 'step' way of defining a sweep so even though I think they should in general be replaced in wrappers I would possibly keep them around somewhere for personal use. I also don't think this is a good use of time to dig through and replace right now because I think we should work out what sort of structure we want the wrappers to have etc and then most of the code in 'transmon' will get refactored or deleted. Talk on Wednesday?

Dominik-Vogel:
Thank you @nataliejpg for pointing out the different signature with step vs num_points. I think specifying a step size could be integrated into doNd without breaking compatibility.
And I also agree, that it is not the time to do all these replacements right now. I made this PR because I found these redundant functions when going through the wrappers in preparation for tomorrow....

nataliejpg:
@Dominik-Vogel oooh that sounds interesting with doNd. Great that you are looking through and seeing what there is. I have the feeling that it will get pulled apart so much that it will basically be rewritten (luckily I'm not very attached to any of the code) but I guess we'll see.  Happy new year to you too

nataliejpg:
I think that deleting the whole folder a la #156 makes more sense. @Dominik-Vogel  are you fine for me to close this?

PR #57, opened 2018-02-07
Fix log level

ThorvaldLarsen:


ThorvaldLarsen:
@Dominik-Vogel What is the correct log level for normal operation?

jenshnielsen:
The default level for console output is warning if you don't do anything else.

ThorvaldLarsen:
Okay, guess we should leave it at WARNING here also. Not sure if we are somehow specifying the log level without needing to.

jenshnielsen:
So Normally if you don't do anything there will be a logger to the screen/console with level warning but we want data to be logged to a file too. Therefore we setup a file based logger with level info but we do want to continue showing the output at level warning to the console so we set it up again. The level for the console is controlled by this config

Dominik-Vogel:
Ok thank you @ThorvaldLarsen , I nobody looked into the config file as the setting has not been read before...

PR #59, opened 2018-02-16
Plotting multiple dataset together with show_num

ThorvaldLarsen:
New functionality for show_num function.

This has not been tested yet - just wanted to make the PR.
Again not the prettiest code and a point on our list for wrappers is to redo this and the show_meta function at some point.


ThorvaldLarsen:
This has now been tested and is working nicely at T11. T3 might also start using it over the weekend.

I have not tested it on 2D datasets not originating from a do2d(). Would be good to test that these datasets have the same structure for calculating x/y limits.

ThorvaldLarsen:
Turns out QtPlot cannot plot a list of datasets while MatPlot can. So for now limited the multiple dataset plotting feature to MatPlot only.

ThorvaldLarsen:
@Dominik-Vogel @nataliejpg Any of you against this PR? Otherwise I would like to merge it.
This has been tested thoroughly on T11. It has also been tested on T4 for do2d datasets.

nataliejpg:
nothing from me, I would have preferred to have it as `shownum(1, 15, 18)` rather than `shownum([1, 15, 18])` but it's such a minor thing I'd say push this and we can always tweak later.

ThorvaldLarsen:
I dont know how to write a function such that shownum(1, 15, 18) would work for arbitrary number of datasets...

nataliejpg:
then leave it and just merge this :)

PR #67, opened 2018-02-28
Feat/logging module

Dominik-Vogel:
This is the first step towards disentangling the wrappers into modules. This module does not depend on anything else in the wrappers.
To use it simply add
```
from qdev_wrappers import logging
```
as the first line of your code. Log files will be created in a central directory (data directory from qcodesrc.json file/logs)

### DIFF
--- a/qdev_wrappers/__init__.py
+++ b/qdev_wrappers/__init__.py
@@ -1,3 +1,4 @@
+from qdev_wrappers import logging
 from qdev_wrappers.file_setup import my_init
 from qdev_wrappers.device_annotator.device_image import save_device_image
 from qdev_wrappers.show_num import show_num
--- a/qdev_wrappers/file_setup.py
+++ b/qdev_wrappers/file_setup.py
@@ -92,48 +92,6 @@ def _init_device_image(station):
     log.info('device image initialised')
 
 
-def _set_up_ipython_logging():
-    ipython = get_ipython()
-    # turn on logging only if in ipython
-    # else crash and burn
-    if ipython is None:
-        raise RuntimeWarning("History can't be saved. "
-                             "-Refusing to proceed (use IPython/jupyter)")
-    else:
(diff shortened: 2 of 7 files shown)


jenshnielsen:
Looks good but if we merge this its going to break the local init script for all users until they remove the original imports

Dominik-Vogel:
This is true... my plan was to take a tour around all the stations and make sure that logging works for every single one. Maybe I can merge the PR on every station locally and then we merge it into master.

jenshnielsen:
I suggest making the old methods aliases for the new ones

Dominik-Vogel:
good point. I put the functions back in with a deprecation warning.

Dominik-Vogel:
It would be great if we could have your PR#66 in here as well

ThorvaldLarsen:
I think we should avoid spending too much time ensuring backward compatibility. My viewpoint is that the current version of wrappers just kinda emerged with a lot of code decisions made on the fly. If this limits the code then I dont see it as a problem that an update breaks previous versions.

To me the challenge is rather to have really good communication channels from developpers to users (that dont know git) which we haven't set up yet. Might be good to have a discussion about this at some point.

Dominik-Vogel:
Thank you @ThorvaldLarsen. This is a very good point. To defend the backwards compatibility here at this point: It was a quick change that took me a few minutes to implement and even though it might not be crucial here, I also learned here how to handle these things for contributions to qcodes core, where it matters.
I haven't set up a qdev-wrappers news channel, because I wanted to finish the new wrapper functionalities to start off with some interesting good news....

ThorvaldLarsen:
@Dominik-Vogel we are starting new experiment on T11 and I am hoping to use the new stationconfig and this logger.
Do you think it would be possible to merge this PR before then? Is anything missing from this PR?

jenshnielsen:
@ThorvaldLarsen Just merged

PR #68, opened 2018-03-02
StationConfigurator

Dominik-Vogel:
In this PR a `StationConfigurator` object is introduced, which enables loading instrument through a yaml file. An example code could look like this:
```


scfg = StationConfigurator('exampleConfig.yaml')
dmm1 = scfg.load_instrument('dmm1')
mock_dac = scfg.load_instrument('mock_dac')
```
With  a yaml file containing:
```
instruments:
  dmm1:
    driver: qcodes.instrument_drivers.agilent.Agilent_34400A
    type: Agilent_34400A
    auto_reconnect: true
    address: GPIB::1::65535::INSTR
    init:
      visalib: 'Agilent_34400A.yaml@sim'
    parameters:
      volt: {monitor: true}
  mock_dac:
    driver: qcodes.tests.instrument_mocks
    type: DummyInstrument
    auto_reconnect: true
    init:
      gates: {"ch1", "ch2"}
```
For the complete example and an extensive commented yaml file take a look at the example in the example folder. For a full list of features refer to the yaml file.
This PR depends on QCoDeS/qcodes#986 

## Current issues:
- [x] Monitor restart still throws error
- [x] Parameters get added multiple times to the monitor if the instrument is reinstantiated.
- [x] Only add config once to metadata
- [ ] Test if it now works with pre/post parsers
- [ ] Make Decadac compatible: init in decadac_ext to set validators, access channels on top level
- [x] Remove entries from station snapshot when reevaluating
- [x] make kwargs being passed on
- [x] find better names
- [x] address on top level
- [x] need to pyyaml to dependencies 
- [ ] make parameter type customizable
- [ ] make standard path for yaml files available


qdev_wrappers/station_configurator.py
```diff
@@ -155,14 +147,12 @@ def setup_parameter_from_dict(p, options):
 
         # add the instrument to the station
         self.station.add_component(instr)
+
+        # restart the monitor to apply the changes
```
jenshnielsen: Is this restart really needed anymore? Creating a new monitor should do the job?

jenshnielsen:
Need https://github.com/QCoDeS/Qcodes/pull/994 and https://github.com/QCoDeS/Qcodes/pull/986

There is a experiments.db file committed that should probably be deleted

qdev_wrappers/station_configurator.py
```diff
@@ -0,0 +1,153 @@
...
+from qcodes.instrument.parameter import Parameter
+from qcodes.monitor.monitor import Monitor
+from .parameters import DelegateParameter
+
+log = logging.getLogger(__name__)
+
+# config from the qcodesrc.json file (working directory, home or qcodes dir)
+auto_reconnect_instrument = qcodes.config["user"]["auto_reconnect_instrument"]
+
+
```
jenshnielsen: Should the atexit register logic be added to the station Configurator? Since it creates the instruments I think so

Dominik-Vogel: I would rather prefer not to. The way it is now the StationConfigurator is an optional component, that you can use if you like it. The atexit should not be optional. But I have to say that I don't have a better idea where to put it.
What about the init of qdev wrappers. It might get added multiple times then but this should not break anything. Now that we are excluding packages from the UMR this should work, right?

jenshnielsen: Yes running it multiple times should be fine. We still need to merge https://github.com/QCoDeS/Qcodes/pull/1009 to fix the UMR issue

jenshnielsen:
@Dominik-Vogel Could you update the example yml for the test setup to the new style too

jenshnielsen:
@Dominik-Vogel Somehow a non standard char ended up on the last line of the testsetup config. Confusing git and pyyaml. I have pushed a fix to remove that

ThorvaldLarsen:
@Dominik-Vogel can we call force_close_existing_instrument for enable_forced_reconnect? Having 'enable' in the name emphasizes that the setting allows _you_ to reconnect instruments not that the code does anything on its own.

Dominik-Vogel:
@jenshnielsen Had I missed anything in the example yaml for the test setup? What do you want me to update?

qdev_wrappers/station_configurator.py
```diff
@@ -0,0 +1,161 @@
...
+        if instr_cfg.get('enable_forced_reconnect',
+                         enable_forced_reconnect):
+            # save close instrument and remove from monitor list
+            with suppress(KeyError):
+                instr = Instrument.find_instrument(identifier)
+                # remove parameters related to this instrument from the monitor list
+                self.monitor_parameters = {k:v for k,v in self.monitor_parameters.items() if v.root_instrument is not instr}
+                instr.close()
+                # remove instrument from station snapshot
+                self.station.components.pop(instr.name)
```
ThorvaldLarsen: If understand this correctly then if `enable_forced_reconnect` is false then the configure instead creates a new instrument in the station.
Is this the behavior we want? I initially assumed the configure would throw an error message and do nothing if `enable_forced_reconnect` is false. Maybe we can add a warning message printed to the console to notify the user that a new instrument instance is created?

setup.py
```diff
@@ -20,7 +20,8 @@
     install_requires=[
         'matplotlib>=2.0.2',
         'pyqtgraph>=0.10.0',
-        'qcodes>=0.1.3'
+        'qcodes>=0.1.3',
```
jenshnielsen: This should probably be bumped to the latest release?

qdev_wrappers/station_configurator.py
jenshnielsen: I think it's going to fail earlier because it's trying to create a new instrument with the same name

Dominik-Vogel: It's good to see that you already got so much used to the StationConfigurator, that you forgot how it was like before ;-). When you try to create the instrument again, you will get an error from qcodes saying that you can't create an instrument with the same name.

Dominik-Vogel: the comments crossed...

ThorvaldLarsen: Ahh, I see. Thanks.

ThorvaldLarsen:
One more request: Would it be possible to add Offset to the list of PARAMETER_ATTRIBUTES. I have a specific instrument that really could benefit from this. 

I realise that this is probably more a question for parameter class in core qcodes but if it is not too hard to add I think it would be a nice feature. This would allow the user to easily adjust for calibration errors on instruments (like a constant voltage offset that is often seen on decadacs),

Dominik-Vogel:
You are totally right, and this is already on my list of things to implement.

ThorvaldLarsen:
@Dominik-Vogel can you add a function for updating the monitor? That is running .get() on all parameters in the monitor list.
Right now I am doing:
`for i in scfg.monitor_parameters: scfg.monitor_parameters[i].get();` but it is a bit confusing figuring out that this works.
Would be nice to just write something like `scfg.monitor_update()` which everyone can understand.

PR #70, opened 2018-03-20
Atexit in init

Dominik-Vogel:
adding Instrument.close_all() and UMR to init.
works with Ipython and through spyder.

### DIFF
--- a/qdev_wrappers/__init__.py
+++ b/qdev_wrappers/__init__.py
@@ -1,5 +1,3 @@
-import atexit
-
 from qdev_wrappers.file_setup import my_init
 from qdev_wrappers.device_annotator.device_image import save_device_image
 from qdev_wrappers.show_num import show_num
@@ -10,4 +8,3 @@ from qcodes.instrument.base import Instrument
 from qcodes.utils.helpers import add_to_spyder_UMR_excludelist
 
 add_to_spyder_UMR_excludelist('qdev_wrappers')
-atexit.register(Instrument.close_all)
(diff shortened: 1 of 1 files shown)


jenshnielsen:
This will need to wait for https://github.com/QCoDeS/Qcodes/pull/1009 otherwise :+1:

Dominik-Vogel:
Hang on, I think this should better go into core... what do you think @jenshnielsen ?

jenshnielsen:
Yes I think so

Dominik-Vogel:
moved atexit part to 
https://github.com/QCoDeS/Qcodes/pull/1017
left UMR

qdev_wrappers/__init__.py
```diff
@@ -2,4 +2,9 @@
 from qdev_wrappers.device_annotator.device_image import save_device_image
 from qdev_wrappers.show_num import show_num
 from qdev_wrappers.sweep_functions import do0d, do1d, do2d, do1dDiagonal
+
 from qcodes.monitor.monitor import Monitor
+from qcodes.instrument.base import Instrument
```
jenshnielsen: I think this is unused now?

jenshnielsen:
Should we wrap this in a try, except Import error to support old version of qcodes? Otherwise :+1:

qdev_wrappers/__init__.py
Dominik-Vogel: This might break things if people are not importing the instrument correctly...

jenshnielsen: Ok fine to leave it

PR #72, opened 2018-03-22
Automated installation

Dominik-Vogel:
This script installs all you need to use qcodes:
- Miniconda (Python), including spyder and jupyter
- Git (version control)
- Qcodes, qdev-wrappers, broadbean
- creates links on desktop and quick access

To make sure everything works as expected go to "Start->Add or remove programs"
and search for "Python" to remove any previous installation of Anaconda,
Miniconda or similiar. Beware that this removes all your python environments
and if you have made changes to any distribution (which you should never do)
you will lose these changes. To see if you have made changes you can either
call "git status" in the corresponding repository directory from the git bassh
or install the github gui.
For a clean reinstallation also remove previous installs of git.
To run this installation script, right-click it and select run with PowerShell


ThorvaldLarsen:
Really good idea! 

One change I make all the time for the qcodes environment is to add scipy packages for fitting. Any chance this can be done automatically?

What kind of git program is installed? I like the current setup with GitHub desktop.

Dominik-Vogel:
I am glad you like it @ThorvaldLarsen :-). At the moment the github gui is not included, but I am happy to add it, or invite you to do it. I suggest we add scipy to the wrappers dependencies.

PR #74, opened 2018-04-09
fix bug in analysis_plot_helpers save_fig function

nataliejpg:
Changes the logic for generating name to save analysis plots under. This use to require the provision of the dataset number that the analysis came from and would break if you didn't provide one. You can now instead provide a custom name. Literally just fixes something broken which is only used by my own automated analysis plotting.


nataliejpg:
@ThorvaldLarsen noone is even using this right? It's buried deep within transmon and is basically the function to save a matplotlib figure you build yourself into the analysis folder. Is it fine if I just go ahead and merge?

ThorvaldLarsen:
not sure what you mean by I build it? But it seems fine to merge this.

nataliejpg:
@ThorvaldLarsen cool, thanks. I just meant if you create a fig and want to save it and it isn't wrapped in qcodes (and hence has a save function already) but it's not important.

PR #76, opened 2018-04-10
rename logging to logger to resolve naming conflict with python logging

Dominik-Vogel:


jenshnielsen:
You need to change file_setup and the moved init file too I think

Dominik-Vogel:
thanks for pointing that out

jenshnielsen:
qdev-wrappers/qdev_wrappers/logger/__init__.py still needs to be updated

jenshnielsen:
I also need an explicit import of logging.handlers in logging_functions.py to make this run

PR #79, opened 2018-04-10
Group Set Parameter

Dominik-Vogel:
This PR introduces a new Parameter that wraps the setting of a group of parameters, as can be understood from the following example:
```
qdac = QDac('qdac', address='ASRL4::INSTR', update_currents=False)
p = GroupSetParameter('bound_channels', qdac.ch01.v, qdac.ch02.v)
p(1)
assert qdac.ch01.v() == 1
assert qdac.ch02.v() == 1
```
It is questionable if this feature should be implemented in form of a parameter. Let's discuss this here:
Pros:
 - can provide new name and label
 - in the new DataSet the parameter can be stored as one parameter (one should store as metadata what parameters it actually refers to.
- additional validators can be set

Cons:
 - as metadata it shows up multiple times (maybe this is not so bad...)
 - it implements all sorts of instrument specific functionality (Let's make a lightweight Parameter class for this!)
     - Command processing
     - valmapping
 - when sweeping with a certain rate this will cause double sweeping rate (this can be implemented in a smarter way, while keeping the concept of a parameter)

This brief discussion is very premature, but I have seen a lot of bad implementations for this functionality, so I wanted to have this somewhat nicer way of doing things in the qdev-wrappers.

### DIFF
--- a/qdev_wrappers/parameters.py
+++ b/qdev_wrappers/parameters.py
@@ -1,0 +1,26 @@
+from qcodes.instrument.parameter import Parameter
+
+class GroupSetParameter(Parameter):
+
+    def __init__(self, name, *parameters, **base_kwargs):
+        if not all(isinstance(parameter, Parameter)
+                   for parameter in parameters):
+            raise RuntimeError("All supplied parameters must be of type Parameter")
+        # include multiparameter and arrayparameter case at some point
+        if len(set(parameter.unit
+                   for parameter in parameters)) > 1:
+            raise RuntimeError("All supplied parameters must have the same unit")
+        if not all(hasattr(parameter, 'set') for parameter in parameters):
+            raise RuntimeError("All supplied parameters must have a set method")
+
+        # construct label if not supplied
+        if 'label' not in base_kwargs:
+            labels = (parameter.label
+                    for parameter in parameters
+                    if parameter.label is not None)
+            if not len(labels) == 0:
+                base_kwargs['label'] = 'Group of {}'.format(labels[0])
+            
+
+        self._parameters = parameters
+
--- a/qdev_wrappers/parameters.py
+++ b/qdev_wrappers/parameters.py
@@ -1,12 +1,49 @@
-from qcodes.instrument.parameter import Parameter
+from typing import Callable
+
+from qcodes.instrument.parameter import Parameter, MultiParameter
+from qcodes.utils import validators
+
 
 class GroupSetParameter(Parameter):
+    """
+    Group a number of parameters to be set simultaneaously.
+    >>> p = GroupSetParameter('bound_channels', qdac.ch01.v, qdac.ch02.v)
+    >>> p(1)
+    here qdac.ch01.v and qdac.ch02.v will both set to 1V
 
-    def __init__(self, name, *parameters, **base_kwargs):
+    Beware that setting a stepping does apply to this parameter
+    and is not inherited to the group of parameters.
+
+    max_val_age not supported
+    """
+    def __init__(self,
+                 name: str,
+                 *parameters,
+                 pre_set: Callable=None,
+                 post_set: Callable=None,
+                 **base_kwargs) -> None:
+        # this class has a different footprint that the Parameter base class
+        # because here the args must be given as kwargs.
+        # this is fine for the Instrument.add_parameter method as it uses
+        # kwargs.
+
+        # type checking for compatibility
         if not all(isinstance(parameter, Parameter)
                    for parameter in parameters):
             raise RuntimeError("All supplied parameters must be of type Parameter")
         # include multiparameter and arrayparameter case at some point
+        if any(isinstance(parameter, MultiParameter)
+               for parameter in parameters):
+            raise RuntimeError("Multiparameter not yet supported")
+        # donnot allow for set_cmd/get_cmd
+        if 'set_cmd' in base_kwargs:
+            raise RuntimeError("Cannot supply set command to GroupSetParameter, because it is defined by setting the individual parameters. Use pre_set and post_set instead.")
+        if 'get_cmd' in base_kwargs:
+            raise RuntimeError("GroupSetParameter is nongettable.")
+
+        # do not check for argument footprint of set functions
+
+        # check that all units are identical
         if len(set(parameter.unit
                    for parameter in parameters)) > 1:
             raise RuntimeError("All supplied parameters must have the same unit")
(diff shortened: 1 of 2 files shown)


jenshnielsen:
So the specific example that you give can already be done as 

```
qdac.channels[0:1].v(1)
```
This probably only has value if you need to group parameters from multiple instruments

Dominik-Vogel:
yes and for nesting. I think you cannot do this with the decadac

jenshnielsen:
The decadac has two levels of nesting. I think you can access it at both levels i.e.

```
decadac.channels[j:k]
```
and 

```
decadac.slots[i].channels[j:k]
```

Dominik-Vogel:
Yes, sorry, I take that back... It was the other way round, there is no eg. ch07 submodule in the decadac...
An alternative route would be to make ChannelList more flexible to support multiple parent instruments.
We could integrate the complete PR into an extension of the ChannelList and MultichannelParameter

jana-d:
I didn't think of something like qdac.channels[0:1].v(1) before, which is indeed a good solution for some cases. 
But "This probably only has value if you need to group parameters from multiple instruments" : Hmmm.

What if you have a wrapper function to do 1d sweeps and you want to sweep two gates at the same time and have a correct label on your graph? And saved in metadata that you swept two things at the same time?
Or if your code has not been written with having inputs like qdac.channels[0:1] in mind, but rather just one parameter and you would love the combined parameter to take care of setting all corresponding values?
(I am all ears if I am over-complicating things and there is a better solution to what I am doing).
Or if users have aliases for channels and would like write something like
cp = GroupSetParameter(dac.left_wall, dac.right_wall)?

I also like the idea to be able to group parameters based on their purpose. E.g I currently have two double dots and would like to say group1 = <all gates of left double dot> and group2 = <all gates of 
 right double dot> and so on.
Again, if you think this is unnecessary or if there might be a better way of doing this then I am all ears.

qdev_wrappers/parameters.py
```diff
@@ -0,0 +1,83 @@
...
+        # because here the args must be given as kwargs.
+        # this is fine for the Instrument.add_parameter method as it uses
+        # kwargs.
+
+        # type checking for compatibility
+        if not all(isinstance(parameter, Parameter)
+                   for parameter in parameters):
+            raise RuntimeError("All supplied parameters must be of type Parameter")
+        # include multiparameter and arrayparameter case at some point
+        if any(isinstance(parameter, MultiParameter)
```
jenshnielsen: Can we even reach this code path. AFAIK MultiParameter is not a subclass of Parameter (but of baseparameter)

Dominik-Vogel:
I agree, with both of you. I had Janas perspective in mind when writing that Parameter: I just wanted to get as much ease of use as possible when using it in the day to day experminets. That means I want to hide possible sources of errors like having to remember channel numbers rather than their alias. 
It is, however, worthwhile thinking about whether one could find a common implementatnion with the ChannelList. 
Some of the things you, @jana-d, are saying can be done with the ChannelList (we all only learned that yesterday...) as the index operation [] return as ChannelListParameter, which can be labeled etc.
If one were to implement a function that makes a slice from a list of parameters than
```
p = qdac.channels[slice_from_list(cutter_right, cutter_left)].v
```
would be very close to 
```
p = GroupSetParameter(qdac.cutter_left, qdac.cutter_right)
```
Another aspect we were discussing is, that some devices implement their own version of a ChannelList through subclassing it, where they override the setting functionality for multiple channels, such that they all get set with a single VISA command. This is a really nice feature, which we don't want to loose.
I think we should look at what a more general implemenation of the channelist would look like.

qdev_wrappers/parameters.py
Dominik-Vogel: Well spotted, this has to be _BaseParameter of course

jana-d:
Sounds all good to me :). 
I think I need to know more about the ChannelList to see if it will work with what I had in mind for automated tuning. 
Thanks!! :)

sohailc:
This is how we solved the problem with pysweep in Delft: 

```python
@getter([('I', ''), ('Q', ''), ('Abs', '')])
def measure_IQ():
    data = iq_acq.do_acquisition()
    data = data.mean()
    result = data.real.astype(float), data.imag.astype(float), np.abs(data).astype(float)
    return result

iq_acq.acq_time = 98e-6
iq_acq.demod_frq = 20e6
iq_acq.update_acquisitionkwargs(records_per_buffer=5000, 
                                buffers_per_acquisition=1,
                                buffer_timeout=10000,
                                allocated_buffers=1)


frqs = np.linspace(7.14e9, 7.16e9, 201)

setup = [(lambda: None, tuple())]
cleanup = [(lambda: None, tuple())]

result = do_experiment(
    "ref-transmon/resonator-spec",
    setup,
    nest(
        szip(
            sweep(LO.frequency, frqs + 20e6),
            sweep(RF.frequency, frqs),
        ),
        measure_IQ
    ),
    cleanup,
    return_format=["data_set_path", "dataset"],
    live_plot_axes=[{'RF_frequency': 1, 'Abs' : 4},]
)
```

This is actual measurement code used in real productive measurements

Dominik-Vogel:
@sohailc Thank you for joining in here! As I am not using the sweep module every day, I have trouble to see how your code snippet is related to this PR. Could you explain how you are grouping the parameters and discuss the advandages and disadvantages over the other two proposed solutions?

jana-d:
Hello again! 
Here is an example of what I had in mind with "parameter groups" or tags (e.g in the .yaml station config file).
I don't necessarily need to set them at the same time, but I would like to know whether they act together on something, e.g a dot:


LW1_1 = Gate(qcodes_parameter=mdac.LW1_1)
LP1_1 = Gate(qcodes_parameter=mdac.LP1_1)
C1_1 = Gate(qcodes_parameter=mdac.C1_1)
RP1_1 = Gate(qcodes_parameter=mdac.RP1_1)
RW1_1 = Gate(qcodes_parameter=mdac.RW1_1)
N1_1 = Gate(qcodes_parameter=mdac.N1_1)

left_double_dot = [LW1_1, LP1_1, C1_1, RP1_1, RW1_1, N1_1]

# left sensing got
LSD_T_1 = Gate(qcodes_parameter=mdac.LSD_T_1)
LSD_C_1 = Gate(qcodes_parameter=mdac.LSD_C_1)
LSD_B_1 = Gate(qcodes_parameter=mdac.LSD_B_1)

left_sensing_dot = [LSD_T_1, LSD_C_1, LSD_B_1]

# right double dot
LW2_1 = Gate(qcodes_parameter=mdac.LW2_1)
LP2_1 = Gate(qcodes_parameter=mdac.LP2_1)
C2_1 = Gate(qcodes_parameter=mdac.C2_1)
RP2_1 = Gate(qcodes_parameter=mdac.RP2_1)
RW2_1 = Gate(qcodes_parameter=mdac.RW2_1)
N2_1 = Gate(qcodes_parameter=mdac.N2_1)

right_double_dot = [LW2_1, LP2_1, C2_1, RP2_1, RW2_1, N2_1]

# right sensing dot
RSD_T_1 = Gate(qcodes_parameter=mdac.RSD_T_1)
RSD_C_1 = Gate(qcodes_parameter=mdac.RSD_C_1)
RSD_B_1 = Gate(qcodes_parameter=mdac.RSD_B_1)

right_sensing_dot = [RSD_T_1, RSD_C_1, RSD_B_1]

#jellybean dot
JBL_1 = Gate(qcodes_parameter=mdac.JBL_1)
JBC_1 = Gate(qcodes_parameter=mdac.JBC_1)
JBR_1 = Gate(qcodes_parameter=mdac.JBR_1)
JBB_1 = Gate(qcodes_parameter=mdac.JBB_1)

jellybean_dot = [JBL_1, JBC_1, JBR_1, JBB_1]



gate_groups = {
    'left_double_dot': left_double_dot,
    'left_sensing_dot': left_sensing_dot,
    'right_double_dot': right_double_dot,
    'right_sensing_dot': right_sensing_dot,
    'jellybean_dot' : jellybean_dot, 
}

This is just a quick setup I implemented, which can probably be more elegant. But during (automated) tuning it would be more efficient to know which gates are actually meant to work together.

Dominik-Vogel:
@jana-d  what does `Gate` do in this case?

jana-d:
@Dominik-Vogel not much yet, it is supposed to hold information I am going to extract. E.g valid voltage ranges etc

PR #81, opened 2018-04-17
Fix dictionary in StationConfig

ThorvaldLarsen:
`for k, v in kwargs` throws an error about expected number of values as only keys are retuned when calling `in kwargs`.
With this fix the init file runs with settings passed through `kwargs` on T3.


ThorvaldLarsen:
Actually this still doesn't work with Alazar...

jenshnielsen:
@ThorvaldLarsen This will not work with the alazar due to the config function requirement until https://github.com/QCoDeS/Qcodes/pull/1003 is merged

ThorvaldLarsen:
so it works by doing using `init_kwargs_all = {**init_kwargs,**kwargs}` instead.

Dominik-Vogel:
Sorry my bad... I just missed the `items()` ....

ThorvaldLarsen:
So this still doesn't actually work with the alazar. It has the same problem as the .update version. On T3 I had to go back to `init_kwargs_all = {**init_kwargs,**kwargs}` as originally...

PR #82, opened 2018-04-19
pass the name explicitly passing of name

jenshnielsen:
This will work even if the config is used for a channel where name is not the first arg 
@Dominik-Vogel


jenshnielsen:
@ThorvaldLarsen We think this will solve your channel in yaml file issue without changing the order of the arguments, which would break other code

ThorvaldLarsen:
Great. Thanks!

PR #84, opened 2018-04-19
reload config file on every load instrument

Dominik-Vogel:
This is a bit slower, but makes it easier to use the SC. Now for changes in the config file for a certain instrument, only the load_instrument line has to be reexecuted.


WilliamHPNielsen:
I like this idea. How much slower is "a bit slower". Is anyone even going to notice?

Dominik-Vogel:
I guess not, because you don't do it programatically... I think it is not worth the effort to benchmark.

WilliamHPNielsen:
Well, then.

Issue #88, opened 2018-04-23
Png save fails for sweep with many measurements (~8 or more)


ThorvaldLarsen:
```
2018-04-22 18:58:28,331 - qdev_wrappers.sweep_functions - ERROR - Exception in doND
Traceback (most recent call last):
 File "C:\Users\t3user\qdev-wrappers\qdev_wrappers\sweep_functions.py", line 167, in _do_measurement
   _do_MatPlot(data,meas_params)
 File "C:\Users\t3user\qdev-wrappers\qdev_wrappers\sweep_functions.py", line 188, in _do_MatPlot
   plot.fig.tight_layout(pad=3)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\figure.py", line 1755, in tight_layout
   self.subplots_adjust(**kwargs)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\figure.py", line 1633, in subplots_adjust
   ax.update_params()
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\axes\_subplots.py", line 114, in update_params
   return_all=True)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\gridspec.py", line 429, in get_position
   gridspec.get_grid_positions(fig)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\gridspec.py", line 90, in get_grid_positions
   subplot_params = self.get_subplot_params(fig)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\gridspec.py", line 376, in get_subplot_params
   hspace=hspace)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\figure.py", line 193, in __init__
   self.update(left, bottom, right, top, wspace, hspace)
 File "C:\Users\t3user\Anaconda3\envs\qcodes\lib\site-packages\matplotlib\figure.py", line 228, in update
   raise ValueError('left cannot be >= right')
ValueError: left cannot be >= right
```

ThorvaldLarsen:
I think it is due to a scalling of the figure size. Maybe it does not take colorbar into account when deciding the size of the figure.

ThorvaldLarsen:
[...]

PR #89, opened 2018-04-23
Hardcode png folder in show_num

ThorvaldLarsen:
This is pretty much a workaraound to avoid #88. Saving of png is done depending on whether 'png_folder' is defined in CURRENT_EXPERIMENT. So to not save as png, which doesn't work, I remove that item in CURRENT_EXPERIMENT, however this currently breaks show_num saving.

Hardcoding the png folder fixes show_num in this case.


Dominik-Vogel:
I would suggest using the qcodes config for these settings. I agree that this should not go into the CURRENT_EXPERIMENT. I want to move all those options into the qcodes config in the long run.

ThorvaldLarsen:
I agree on the long term this should definitely be in the config file somewhere instead. However, shortterm this allows me to do Liveplotting without breaking show_num for png saves.

ThorvaldLarsen:
Solution found to #88 so closing this PR.

PR #90, opened 2018-04-23
Copy initkwargs before modifying them

jenshnielsen:
@Dominik-Vogel and @ThorvaldLarsen 

Fixes issues with non picklable alazar paramaters


Dominik-Vogel:
Thank you for identifying the error!

PR #92, opened 2018-04-24
Added StationConfigurator port setting

emartinez:


ThorvaldLarsen:
@Dominik-Vogel I guess this doesn't matter too much but I think this makes the code more confusing rather than better. Should there be a special case handling all parameters for init settings?

Dominik-Vogel:
@ThorvaldLarsen I understand your concern, as this is not 100% stringent, but instead creates comfort. If you ask me, the address and port take a special role. The address is required for almost all instruments and often the only argument in init. This is why it can also specified in the elevated level. It is consequent to also do this with the port, as it can be seen as equivalent to the address.  All other parameters are either optional or instrument specific.

PR #99, opened 2018-05-14
Znb

ThorvaldLarsen:
Move `self.channels.autoscale()` inside `if` statements so that it is only called if channels are created. Previously autoscale would throw an error if ZNB is initialized with no channels.

@Dominik-Vogel can you take a quick look? Should be fast to merge.

### DIFF
--- a/qdev_wrappers/customised_instruments/ZNB_ext.py
+++ b/qdev_wrappers/customised_instruments/ZNB_ext.py
@@ -72,13 +72,14 @@ class ZNB_ext(ZNB):
 
         if S21:
             self.add_channel(channel_name='S21')
+            self.channels.autoscale()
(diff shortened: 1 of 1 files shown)


Dominik-Vogel:
I would prefer a simpler approach:
leave it where it is and check if it has channels. This makes it safer if we are going to implement other types of channels.

ThorvaldLarsen:
Not sure what you mean by this.

ThorvaldLarsen:
@Dominik-Vogel actually thinking more about it I think we should just completely remove it. The VNA is initialized with RF power off and running autoscale with power off doesn't really make sense.

Dominik-Vogel:
ok that kinda makes sense,

PR #100, opened 2018-05-16
New Alazar_ext driver

ThorvaldLarsen:
This is the new Alazar_ext driver for use with the new Alazar driver with standard parameters (https://github.com/QCoDeS/Qcodes/pull/1003) and moving settings to the yaml file (see example in #83 ).

Also removed the controller part as we will be using the new channels instead.

Both #83 and #1003 should be merged before this one. However, this PR has been tested and works at T3 so should be merged right after.

@nataliejpg @Dominik-Vogel this one removes the previous _ext driver instead of making a new one. I think it is better to just remove the old one instead of letting it sit around without anyone using it.


nataliejpg:
So long as noone else who is using the old one has a problem I'm fine with this. Just good that you gave a heads up so I know I need to change the scripts before I start using this.

PR #103, opened 2018-05-17
[WIP] closing all but the latest 10 matplotlib figures on figure creation to avoid out of memory crashes

Dominik-Vogel:


jenshnielsen:
I think 10 is a bit on the low side, perhaps this should use the same config variable as pyqtgraph

Dominik-Vogel:
Figures get all closed correctly and `get_fignums()` returns an empty list. However memory usage still increases. It seems like there is another data leak somewhere. At @jenshnielsen do you agree that this most probably originates from Matplotlib?
An explicit call to `gc.collect()` does not bring any changes.

Dominik-Vogel:
There is no dataleak when plotting is deactivated via 
````
CURRENT_EXPERIMENT['pdf_subfolder'] = None
CURRENT_EXPERIMENT['png_subfolder'] = None
```
and neither if only pdfs are generated.

Dominik-Vogel:
This lead me to believe that there is a problem with the matplotlib backend.
Switching from Spyders default *module://ipykernel.pylab.backend_inline* to *Qt5Agg* solves the problem

Dominik-Vogel:
Seems to be a know but unfixed issue:
https://github.com/ipython/ipython/issues/7270

PR #107, opened 2018-05-29
new fitter for wrappers

lillian542:
In progress extension to fitting compatible with new dataset.
includes
- [ ] dataconverter (from dataset to dict for both old and new datasets)
- [ ] fitter (based on @ThorvaldLarsen qdev_fitter.py) to fit to dict spat out of dataconverter which itself spits out dict of fit params
- [ ] plotting for fits based on dict spat out by fitter (and doesn't use any qcodes or qdev wrappers)
- [ ] dataconverter (from dict spat out by fitter into new dataset which refs initial dataset and stores fit params and fit function etc)
- [ ] plotter which can plot data from a dataset created by the step above, uses same plotting as qcodes

works for 1d and 2d data, takes cuts or fits to whole 2d
 
@Dominik-Vogel @nataliejpg @ThorvaldLarsen


qdev_wrappers/fitting/AnalysisToSQL.py
```diff
@@ -0,0 +1,151 @@
...
+
+    tablename = 'analysis_{}_{}'.format(fit['inferred_from']['run_id'], fitclass.name)
+    
+
+    table_columns = [est_name]
+    for parameter in fitclass.p_labels:
+        table_columns.append(parameter)
+
+    table_rows = []
+    if dim == 1:
```
Dominik-Vogel: To me the two bodies of the if statements look identical. (I might be overlooking the essential part) If they are equal it would be more readable with if ... or...

qdev_wrappers/fitting/Converter.py
```diff
@@ -0,0 +1,236 @@
...
+    if datatype == 'qcodes_legacy':
+        converter = Legacy_Converter()
+
+    if datatype == 'SQL':
+        converter = SQL_Converter()
+
+    return converter.convert(id, samplefolder)
+
+
+class DataConverter():
```
Dominik-Vogel: It would be great if you could define the used functions and let them throw an exceptions to mark that they are not implemented in this base class. (This would be the classical oop way to do things. Possibly it is fine in Python not to do this, but then I would add a comment that this class is meant to be used as a base class for ... and ... or call it a mixin class.)

qdev_wrappers/fitting/AnalysisToSQL.py
```diff
@@ -0,0 +1,151 @@
...
+def is_table(tablename, cursor):  # Checks if table already exists. Assumes database connection and cursor already established
+
+    table_count = "SELECT count(*) FROM sqlite_master WHERE type = 'table' AND name='{}'".format(tablename)
+    execute = cursor.execute(table_count)
+    count = execute.fetchone()[0]
+
+    if count == 0:
+        return False
+
+    if count == 1:
```
Dominik-Vogel: if the `count` is not 0 or 1 the function does not return anything. It would be great to throw an error in this situation.

qdev_wrappers/fitting/AnalysisToSQL.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def make_table(tablename, cursor):
+    name = tablename
+    n = 0
+
+    while is_table(name, cursor):
+        n += 1
+        name = "{}_{}".format(tablename, n)
+    else:
```
Dominik-Vogel: A `while ... else` statement 👍 😃

qdev_wrappers/fitting/AnalysisToSQL.py
```diff
@@ -0,0 +1,151 @@
...
+    est_name = '{}_estimate'.format(fit['estimate']['name'])
+    est_label = fit['estimate']['label']
+    est_unit = fit['estimate']['unit']
+    param_values = fit['estimate']['parameters']
+
+    dim = fit['inferred_from']['dimensions']
+
+    if dim == 1:
+        param_units = [fit['parameters'][param]['unit'] for param in list(fitclass.p_labels)]
+    elif dim == 2:
```
Dominik-Vogel: Also here you could add an exception if `dim > 2`

qdev_wrappers/fitting/AnalysisToSQL.py
```diff
@@ -0,0 +1,151 @@
...
+    exp_id = fit['inferred_from']['exp_id']
+    predicts = est_name.strip("estimate").strip("_")
+    estimator = "{}, {}".format(fit['estimator']['method'], fit['estimator']['type'])
+    analysis = fit['estimator']['function used']
+    dill_obj = fit['estimator']['dill']
+    parameters = fitclass.p_labels
+    param_labels = fitclass.p_names
+
+
+    conn = sqlite3.connect('experiments.db')  # should this go in a separate analysis database, or just go in experiments.db?
```
nataliejpg: I think we've decided it should go into the experiments database but also it should probably not have the name of the database hard coded, i think it should be possible to get it from the qc.config

qdev_wrappers/fitting/AnalysisToSQL.py
```diff
@@ -0,0 +1,151 @@
...
+    while is_table(name, cursor):
+        n += 1
+        name = "{}_{}".format(tablename, n)
+    else:
+        cursor.execute('CREATE TABLE {} (id INTEGER)'.format(name))
+
+    return name
+
+
+def fit_to_SQL(fit):    #it would be an improvement if it were able to get the fitclass from the fit information
```
nataliejpg: I would <3 <3 <3 some more comments in this function, I know that in theory you shouldn't need comments to explain what you are doing but a few here to break up this rather large function and explain the different sections would be great, especially because as it stands I don't know what the slightly monstrous 'fit' contains/needs to contain.  Also a comment at the start of each function to specify the arguments and broadly what is done/returned would be fab.

qdev_wrappers/fitting/Converter.py
```diff
@@ -0,0 +1,236 @@
...
+class SQL_Converter(DataConverter):
+    
+    def find_data(self, id, samplefolder):  
+        
+            '''sample folder does nothing except act as placeholder.                    
+            This function currently proceeds under the assumption that the experiment is set up, and doesn't
+            check anything to confirm that. Also, I think this doesn't return any error if it can't find 
+            the dataset with that id number.'''
+          
+            data = get_data_by_id(id)
```
nataliejpg: does this assume that the database is stored directly in the samplefolder? I can test this if you don't get a chance :)

qdev_wrappers/fitting/Converter.py
```diff
@@ -0,0 +1,236 @@
...
+        for data_subset in data:
+            for variable in data_subset:
+                if variable['name'] not in all_variables:
+                    all_variables.append(variable['name'])
+                    
+        return all_variables
+
+    def find_experiment(self, id):
+
+        conn = connect('experiments.db')  # the name of the file to access can't be set manually right now
```
nataliejpg: again I think this is in the qc.config so should be possible to get it without it being hardcoded. It might also be in CURRENT_EXPERIMENT

qdev_wrappers/fitting/Converter.py
nataliejpg: also would be a fab place to describe what each function should do and then in the classes which inherit you don't need to do it :)

qdev_wrappers/fitting/Fitclasses.py
```diff
@@ -0,0 +1,72 @@
+
+import qcodes as qc
+import numpy as np
+import scipy.fftpack as fftpack
+
+
+class Fitclass:
```
nataliejpg: Would also make sense for this class to have the functions 'fun' and 'guess' which throw the NotImplementedError in the base class and then are implemented in the classes that inherit. This is also maybe a good place for us to say we are doing a least squares fit and then make this a LeastSqauresFit class. This of course only holds if we are only going to use this class together with the 'fitter' (function or class, it's currently a class but we talked about turning it into a function), at the moment seems reasonable to me but perhaps it's a bit weird to specify that here when there is no actual 'least squares fit' going on in this class.

qdev_wrappers/fitting/Fitter.py
```diff
@@ -0,0 +1,289 @@
...
+    if fun_dim == 1 and num_inputs == 1 and num_setpoints == 0:
+        fitter = Fitter_1D(data, fitclass, fun_output)
+    elif fun_dim == 1 and num_inputs == 1 and num_setpoints == 1:
+        fitter.Fitter_2Ddata_1Dfunction(data, fitclass, fun_output)    #what does the fitter actually need??
+    else:
+        print('The specified function has ')
+
+    fitter.find_fit(data, fitclass, fun_inputs, fun_output, setpoint_params, p0, **kwargs)
+
+class Fitter1D:
```
nataliejpg: maybe I need to read this a bit more carefully but I'm wondering if this is really a Fitter1D or really just a class describing the fit which could have more than 1D? Is this just a Fitter? Essentially I'm asking what the 1D refers to? Also it seems a bit confusing to have both a Fitclass and a Fitter class, this may well be the best way but if so then it is definitely a case where a few lines of description on each and a good paragraph in the example notebook would go a long way :)

qdev_wrappers/fitting/Fitter.py
```diff
@@ -0,0 +1,289 @@
...
+        estimate['name'] = '{}_estimate'.format(self.output_dataname[0])
+        estimate['label'] = '{} estimate'.format(data[self.output_dataname[0]]['label'])
+        estimate['unit'] = data[self.output_dataname[0]]['unit']
+        estimate['parameters'] = self.full_parameter_dict   #not sure this is used for anything
+
+        return estimate
+
+
+class Fitter_2Ddata_1Dfunction(Fitter1D):
+
```
nataliejpg: again a comment describing the class would be great, I also wonder if a class is the best way to implement this. I feel like we want to make something general enough that we can say 'fit to these n parameters for each of these m setpoint parameters and give me a dataset with m values for each output param of the fit'. One instance of this is the 2D dataset and a 1D fit but as soon as we have more setpoint parameters in a dataset or are fitting to more than 2 parameters the number of these classes we need will blow up, maybe we could just make one general one and push the specific stuff into the plotting which is anyways more of a 'convenience wrapper'

qdev_wrappers/fitting/Converter.py
Dominik-Vogel: I would prefer going with the qcconfig rather than creating a dependency on CURRENT_EXPERIMENT

qdev_wrappers/fitting/AnalysisToSQL.py
lillian542: This is gone now, replaced by something more general that doesn't need to worry about the number of variables in the fit function.

qdev_wrappers/fitting/AnalysisToSQL.py
lillian542: Oops...

qdev_wrappers/fitting/AnalysisToSQL.py
lillian542: The 'fit' is the fit dictionary, as returned by the Fitter. The comment at the very beginning is misleading because its a relic of a previous version of the function that had multiple arguments. I've deleted it, reorganized the function a bit and added more relevant comments, so I think it should be more readable now.

qdev_wrappers/fitting/AnalysisToSQL.py
lillian542: Also gone now :)

qdev_wrappers/fitting/AnalysisToSQL.py
lillian542: The function now takes an additional optional argument, database, which allows specification of the database location, so its not hard coded anymore. Letting it get the database from qc.config is still a 'ToDo' though.

qdev_wrappers/fitting/Converter.py
lillian542: get_data_by_id uses qc.config to get the database location :)

qdev_wrappers/fitting/AnalysisToSQL.py
lillian542: Actually I think getting the database from qc.config should work too now.

qdev_wrappers/fitting/Fitter.py
lillian542: What makes this specific to 1D is the 'perform_fit' function. I'm not sure how to write that part generically, since it seems to me that the data has to be organized differently depending on the dimensionality of the data and the function. I've changed it to have a general Fitter class as a base, and then two classes that inherit from that, to make it a bit clearer with the function specific to a '1D' fit (i.e. '1D fit to 1D data') or a '1D fit to 2D data'. 

I've gone back and forth a fair bit between it being a class and just being a function. Right now I've landed on a class, because it allows class inheritance if it turns out that there isn't a good way to make perform_fit generic in terms of number of variables. I also couldn't get dill to work for a function, whereas it worked without issue for an instance of a class. I'm not sure what I was doing wrong there.

qdev_wrappers/fitting/Fitter.py
lillian542: I'm not sure how to go about making them all one class, but I agree that that would be optimal. It seems like different reorganization of the data is needed depending on whether there are setpoints, and how many variables act as set points.

ThorvaldLarsen:
Just switched to the new dataset and I am curious how far along this PR is. Is there a working prototype?

PR #109, opened 2018-06-01
Added commit hash to snapshot using the ConfigComponent class.

majoqubitT10:
When (re-)initializing qcodes such that changes in whatever modules are incorporated into the workflow, the class ConfigComponent is called once. 
I think the initialization method of this class is a rather logical point to query the git commit hashes and add them to the metadata. In fact, by adding the information to the  ConfigComponent.data['codebase']  attribute, the information also ends up at a logical point in the metadata dict. 

To query the git commit hashes, we look into qcodes.config.user.codebase, for the folder that contains qcodes related repositories. When there is no such config key, the ConfigComponent.data['codebase'] will simply contain a empty list. 

The location of the git executable can be specified by setting qcodes.config.core.gitpath. When there is no such key, we will fall back to the Git for Window default location.

### DIFF
--- a/qdev_wrappers/file_setup.py
+++ b/qdev_wrappers/file_setup.py
@@ -51,14 +51,17 @@ def close_station(station):
             pass
 
 
-def _set_up_exp_folder(sample_name: str, mainfolder: str= None):
+def _set_up_exp_folder(sample_name: str, mainfolder: str= None, datafolder: str= "", plot_x_position=0.66):
     """
     Args:
-        mainfolder:  base location for the data
-        sample_name:  name of the sample
+        sample_name:  name of the sample (required)
+        mainfolder:  base location for the data. (optional, otherwise pulled from qcodesrc.json)
     """
     if os.path.sep in sample_name:
-        raise TypeError("Use Relative names. That is without {}".format(sep))
+        raise TypeError("Use relative names. That is without {}".format(sep))
+
+    if os.path.sep in datafolder:
+        raise TypeError("Use relative names. That is without {}".format(sep))
 
     if mainfolder is None:
         try:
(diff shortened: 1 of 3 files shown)


ThorvaldLarsen:
Great idea! Would be nice to keep track of the code version being used.
I have two suggestions:
- Move the path setting from qcodes.config.user to the .yaml file. I hope we can move all settings from that file to the yaml file down the line instead of having two config files.
- Can we remove the 'master' branch warning? It happens quite often that we end up running on non-master branch for some time while drivers are developed and seems a bit much to have a warning everytime we restart.

majoqubitT10:
1) I agree that having those two files is a mess.. I will have a look how we can move the path setting in a clean way to the yml file. 
2) I think having the warning is nice, because it is very (too) easy to miss the fact that there is a different branch active than the one you should/want to use. It is just a warning (no error or something) and it is only thrown at you once when you initiate you setup. I thought that would not be too invasive, more like a little reminder you are not running on the master branch. 
We can also add a keyword argument to the StationConfigurator with the intended branch-names. When a different branch is active on initialization, you will get the error. I do think however, that this is a little less nice solution.

ThorvaldLarsen:
I dont feel strongly about the warning - just my initial thought on it. Agree that the key-word solution is less nice - would rather get a warning than keep changing a key word.

Dominik-Vogel:
Putting this kind of information into the yaml file would break the current intended scope of data.
I am not saying that it is the wrong place, just that we need to take a general design decision here.
@ThorvaldLarsen my interpretation, why you would want it in the yaml file is that you feel that the json file is impractical due to necessity of having a schema and its slightly more complicated syntax. Is this correct?
Up to now, all configuration is in the json file except for instrument specific setup. The yaml file takes a different role in the sense that its config can be experiment/run specific and may be relevant for interpreting results (actually the snapshot should cover it all). Everything that is stored in the json file is convenience and user setup. As such the git path should go into the json file.

ThorvaldLarsen:
I dont really have anything against the json file type. My issue is having two files - both are config files containing the same type of information. Both the information in the yaml and the json file are user and setup specific so I dont see a big difference between them. It is true that some of the information in the yaml file is important for the experiment but all this information is already stored on the parameters via the snapshot. So in the end I don't think the yaml file is more important than the json.

In short: Would be nice to only have one file that contains all the settings not defined in the init.py file. Just seems a lot more intuitive to me.

WilliamHPNielsen:
>  both are config files containing the same type of information. 

If I understand Dominik correctly, then that is wrong (should be wrong, in any case). One - the json file - describes your computer's setup, like your file paths and the maximum number of plots you want to retain in memory. The other - the yaml file - describes your instruments' setup, like the IP addresses and voltage ranges.

That one is json and the other yaml is of course without importance, but the seperation between the two seems very sensible to me.

ThorvaldLarsen:
I simply disagree with this distinction of the two files. As I see it the IP address of an instrument and where on the computer data is saved is the same type of information - information added in a separate config file loaded into python via the init file.

It is more intuitive to me to only load from one config file. I don't see what I gain from having two files other than two files that needs to be copied aroud and modified accordingly when switching computer/fridge.

Of course if others disagree we can keep both files. I just dont want it to be due to historical reasons such as the json file came first...

PR #111, opened 2018-06-07
Live plotting and sweep functions

emartinez:
- The sweep functions (do0d, do1d, do2d) work roughly as they used to, plus a progress bar with timing.

- For the live plotting there is a daemon that has to be run once in a stand-alone Python terminal. The daemon will listen to updates to the database, plot them as they come, and save them.

### DIFF
--- a/qdev_wrappers/live_plotting.py
+++ b/qdev_wrappers/live_plotting.py
@@ -1,5 +1,4 @@
 '''This module plots measurement runs in real time.
-
 To start the plotting daemon, run this module on a stand-alone Python terminal.
 It will not work on a Spyder console because of multiprocessing voodoo.'''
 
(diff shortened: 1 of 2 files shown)


nataliejpg:
@Dominik-Vogel are we waiting on any reason not to go ahead and merge this? maybe the name of 'sweep_functions_new' is a bit dodgy but we can always fix it later when everyone is on the new dataset right? @emartinez I assume from your end this is good to go?

nataliejpg:
I guess the main thing missing is that the doNd here do not take tasks, not sure if that's a bad thing or if people should just learn to write their own loops if they want a task executed.

Dominik-Vogel:
I think we should not support tasks with the new doNds. The whole point was to get rid of this construct.
If @emartinez feels it is ready for it, I can merge it.

nataliejpg:
Thanks @GateBuilder. Could you point me toward the current reccomended live plotting software?

PR #112, opened 2018-06-12
StationConfigurator: tab completion for loading instruments from yaml file

Dominik-Vogel:
With this PR methods are added to the station configurator, that are called `load_<instrument_name>` where `instrument_name` is an instrument that is defined in the yaml file. With this you can do
```
sc = StationConfigurator()
# type sc.load_<tab> and tab complete to get
sc.load_AWG5014()
```
also kwargs can be passed on.

### DIFF
--- a/qdev_wrappers/station_configurator.py
+++ b/qdev_wrappers/station_configurator.py
@@ -1,5 +1,6 @@
 from contextlib import suppress
 from typing import Optional
+from functools import partial
 import importlib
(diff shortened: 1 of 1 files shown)


jenshnielsen:
I like it but we may have to tackle that todo before we merge

Dominik-Vogel:
I resolved the todo @jenshnielsen.

PR #114, opened 2018-06-18
Add transpose option for show_num

ThorvaldLarsen:
See title.

### DIFF
--- a/qdev_wrappers/show_num.py
+++ b/qdev_wrappers/show_num.py
@@ -61,7 +61,7 @@ def show_num(ids, samplefolder=None,useQT=False,avg_sub='',do_plots=True,savepng
                 raise ValueError('qcodes.QtPlot does not support multigraph plotting. Set useQT=False to plot multiple datasets.')
             if dataname is not None:
                 if dataname not in [key for key in data.arrays.keys() if "_set" not in key]:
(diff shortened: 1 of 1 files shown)


ThorvaldLarsen:
@Dominik-Vogel when you have time can you take a quick look to see if okay to merge?
The new feature has been tested.

ThorvaldLarsen:
Actually only works when creating one plot. Something goes wrong if more plots are to be created...

Dominik-Vogel:
sure, let me know when you figure out the bug or need assistance in finding it.

ThorvaldLarsen:
Turns out the issue is if several dataset have the same setpoints. That is set_arrays points to the same thing in memory when I then swap the two set_arrays they are swapped for all. This becomes a problem when the loop then swap the same set_points on second run through.
```
data.alazar_ctrl_rec_f2_phase_data.set_arrays[1] is data.alazar_ctrl_rec_f2_mag_data.set_arrays[1]
Out[8]: True
```
For now I have enforced that a dataset name is provided so that only one graph is produced. It still does what we want although a bit anoying to have to run it many times if there are multiple dataset.
However, I dont think it is worth spending too much time on show_num as it will be outdated when we soonish move to the new dataset.

I think it should be fine to merge as it is now.

PR #117, opened 2018-06-27
Extended SR860

ThorvaldLarsen:
Extending SR860 driver to measure conductance and resistance assuming it is used in combination with a current amplifier.

@Dominik-Vogel can you take a look?

### DIFF
--- a/qdev_wrappers/customised_instruments/SR86x_ext.py
+++ b/qdev_wrappers/customised_instruments/SR86x_ext.py
@@ -1,9 +1,6 @@
 from qcodes.instrument_drivers.stanford_research.SR86x import SR86x
 
 class SR86x_ext(SR86x):
-    """
-    The SR860 instrument is almost equal to the SR865, except for the max frequency
-    """
     def __init__(self, name: str, address: str, **kwargs: str) ->None:
         super().__init__(name, address, **kwargs)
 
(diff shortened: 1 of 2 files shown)


jenshnielsen:
Perhaps it would be better to extend the sr86x driver and then make the sr860_ext a subclass of this extension in the same way as done in the regular drivers. This way we can always do the same for the sr865 and sr865a

ThorvaldLarsen:
Hmm, might be a good idea.

ThorvaldLarsen:
@jenshnielsen does it work this way? I would like to avoid having too many files in the customized instrument folder...

qdev_wrappers/customised_instruments/SR86x_ext.py
```diff
@@ -0,0 +1,57 @@
...
+        I = self.X()/self.iv_gain.get_latest()
+        return (V/I)
+
+
+class SR860_ext(SR86x_ext):
+    """
+    The SR860 instrument is almost equal to the SR865, except for the max frequency
+    """
+    def __init__(self, name: str, address: str, reset: bool=False, **kwargs: str) ->None:
+        super().__init__(name, address, max_frequency=5E3, reset=reset, **kwargs)
```
jenshnielsen: Note this is wrong we just merged https://github.com/QCoDeS/Qcodes/pull/1164 to fix that

ThorvaldLarsen: Thanks. My local qcodes version seems to be slightly outdated.

jenshnielsen:
Other than that I think it should be fine

ThorvaldLarsen:
Will test that is still works and then merge it.

PR #119, opened 2018-06-28
Switch to Ruamel in stationconfigurator

### LINKED ISSUE #110: Dublicates in yaml file ignored silently by StationConfig
With many instruments the .yaml file can get quite long and the possibility for mistakes fairly high over time. Recently had two instruments in the .yaml file duplicated without any errors.

Currently the code ignores the first one in the .yaml file and initialises with the last entry of a given name.

I propose that a check is added to the station configurator such that an error is raised if not all entries in the .yaml file have unique names.

@Dominik-Vogel

ThorvaldLarsen:
Still having problems with scientific notation in yaml file (@Dominik-Vogel this works for some instruments if the parameter has get_parser=float unfortunately this is not the case for all instruments).
Ended up spending 5 min looking into ruamel and it works right out of the box now with scientific notation! Also the duplicate error message is extremely well done (tested it on T3 and found I had a duplicate in a set of parameters).

@Dominik-Vogel or @jenshnielsen can you follow up on this? It is tested and works on T3 with full cQED and transport set of instruments. However, this pull request is missing some parts for installing ruamel.yaml instead of pyyaml when installing qdev-wrappers.

Fixes #110.

### DIFF
--- a/qdev_wrappers/station_configurator.py
+++ b/qdev_wrappers/station_configurator.py
@@ -3,7 +3,6 @@ from typing import Optional
 from functools import partial
 import importlib
 import logging
-from ruamel.yaml import YAML
 import os
 from copy import deepcopy
 import qcodes
@@ -13,6 +12,15 @@ import qcodes.utils.validators as validators
 from qcodes.instrument.parameter import Parameter
 from qcodes.monitor.monitor import Monitor
 from .parameters import DelegateParameter
+use_pyyaml = False
+try:
+    from ruamel.yaml import YAML
+except ImportError:
+    import yaml
+    use_pyyaml = True
+    warnings.warn("ruamel yaml not found station configurator is falling back to pyyaml. "
+                           "It's highly recommended to install ruamel.yaml. This fixes issues with "
+                           "scientific notation and duplicate instruments in the YAML file")
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


jenshnielsen:
@ThorvaldLarsen How would you fell about a fallback i.e. try to use ruamel and fallback to pyyaml is it does not import with a warning.

jenshnielsen:
Ok 

I think the following should do the trick

```python
use_pyyaml = False
try:
    from ruamel.yaml import YAML
except ImportError:
    import yaml
    use_pyyaml = True
    warnings.warn("ruamel yaml not found station configurator is falling back to pyyaml. "
                           "It's highly recommended to install ruamel.yaml. This fixes issues with "
                           "scientific notation and duplicate instruments in the YAML file")

...
        def load_file(self, filename: Optional[str] = None):
            if use_pyyaml = False:
                yaml=YAML()
```

ThorvaldLarsen:
Yes, that looks good. I can test it both before and after installation of ruamel on T11 tomorrow.
I also think that ruamel.yaml should go in setup.py at some point rather than pyyaml but I dont feel strongly about that as long as we have the ruamel functionality here.

jenshnielsen:
@ThorvaldLarsen I agree but I think we might hold of with that until we move the configurator into qcodes it self

ThorvaldLarsen:
@jenshnielsen had to move the import yaml part from the except part to get it to work. However, the current version works at T11 both with and without ruamel.yaml installed.

Okay to merge?

jenshnielsen:
Ok from me any objections @Dominik-Vogel

PR #120, opened 2018-06-28
Update SR830_ext for use with SCFG

ThorvaldLarsen:
Update SR830_ext for use with the new stationconfigurator. Also add resistance and conductance measurements (this is a complete copy of the SR86x_ext only on a different subclass).

Still need to test it.

### DIFF
--- a/qdev_wrappers/customised_instruments/SR830_ext.py
+++ b/qdev_wrappers/customised_instruments/SR830_ext.py
@@ -19,7 +19,7 @@ class ConductanceBuffer(ChannelBuffer):
         # If X is not being measured, complain
         if self._instrument.ch1_display() != 'X':
             raise ValueError('Can not return conductance since X is not '
-                             'being measured on channel 1.')
+                                'being measured on channel 1.')
 
         resistance_quantum = 25.818e3  # (Ohm)
         xarray = super().get()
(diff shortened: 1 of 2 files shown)


ThorvaldLarsen:
This has now been tested and been shown to work. I did not test the conductance channel part but this has not really been changed from what it was before.

Walked through lab and nobody is using this driver (it is also old and not compatible with the station configurator) so should be fine to modify it. 
Everyone is instead using manual paramters in the init script to get conductance and resistance. This was done on at least 5 different setups so there is a real need to have this in the driver.

qdev_wrappers/customised_instruments/SR830_ext.py
```diff
@@ -37,81 +34,73 @@ def get(self):
...
-
-        self.add_parameter('resistance',
-                           label='{} Resistance'.format(self.name),
+       	self.add_parameter(name='resistance',
+                           label='Resistance',
+                           unit='Ohm',
                            get_cmd=self._get_resistance,
+                           get_parser=float)
+        
+        self.add_parameter(name='g_X',
```
jenshnielsen: This indentation seems to be off

jenshnielsen: Or rather above, I think some taps have ended up in the code

ThorvaldLarsen: Weird. It still runs though. Tried redoing the indentation but feel like there is an automatic way of doing that which I haven't learned yet...

jenshnielsen: Depending on the editor there is usually a convert taps to space setting and a indentation config

Dominik-Vogel:
@ThorvaldLarsen you added a factor 1/2 for the conductance. Is this not a problem for everybody to suddenly have different results? (I mean have you communicated this? we should somehow capture this in meta data...)

ThorvaldLarsen:
It is already in the metadata with the unit being `e^2/h` instead of `2e^2/h`, so will be obvious. The choice of unit was based what people are using at the moment.

Also, nobody was using the old driver so it will not be a surprise change for anyone. Instead everyone have manual parameters in the init script using the bare qcodes driver. So changing this will not change any code for anyone but rather give a better option than what is currently being done.

Dominik-Vogel:
ok, I just overlooked that you also changed the unit...

PR #122, opened 2018-07-02
Add offset as parameter_attribute in SCFG

ThorvaldLarsen:
@Dominik-Vogel by chance I discovered my issue with `monitor.update_all()`. My problem was I didn't call it from the initialized monitor object. With this change one can do `scfg.monitor.update_all()`.

Also added `offset` to the list of parameter attributes.

### DIFF
--- a/qdev_wrappers/station_configurator.py
+++ b/qdev_wrappers/station_configurator.py
@@ -40,11 +40,12 @@ class StationConfigurator:
     """
 
     PARAMETER_ATTRIBUTES = ['label', 'unit', 'scale', 'inter_delay', 'delay',
-                            'step']
+                            'step', 'offset']
 
     def __init__(self, filename: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


Dominik-Vogel:
So you are saying that with this PR
```
scfg.monitor.update_all()
```
works for you, but:
```
Monitor.running.update_all()
```
does not?

👍 fro adding offset! I totally forgot to put it into the station configurator as well when I added it to the parameter. It is great to see that someone is using the offset!

ThorvaldLarsen:
I did not know about the `Monitor.running` part. Thanks! 

Is there a doc for the monitor somewhere - I could not find one. Maybe a short writeup would be good at some point?

ThorvaldLarsen:
I removed the changes to the monitor part. But guess we can still merge the offset.

Dominik-Vogel:
True, we are lacking documentation. Not only for the monitor.

PR #123, opened 2018-07-09
Update Alazar notebook to use new way of setting parameters

jenshnielsen:
Not using the config object but setting the parameters manually and syncing with the syncing context manager. @Dominik-Vogel


nataliejpg:
shall we merge this...? looks fine to me

jenshnielsen:
Totally forgot about this.

PR #124, opened 2018-07-18
Parametric Sequencer

Dominik-Vogel:
A new restructured parametric sequencer that takes care of sequences that are created by a array of one or a grid of two symbol values applied to a single template element.


ThorvaldLarsen:
Looking forward to this! :)

One suggestion: Would it be possible to think of the number of dimensions as any integer rather then 1 or 2? With plotting tools being developped for handling larger dimensional dataset I think we should avoid thinking of a dataset being limited to 2D.

nataliejpg:
- isn't it bad to put a things like lists and dicts as a default value (http://effbot.org/zone/default-values.htm)? I'm specifically looking at `set_sequencer_template(context: Dict={})`. I can't see how it would make any difference here I'm just very much 'once bitten twice shy' about doing it and wondering if it was deliberate/conscious?

nataliejpg:
- could you maybe give some thoughts on the two different channels sequence and repeat. Specifically I find it confusing that the parameter which is being varied throughout the sequence appears in both and that you can set/get it from the sequence channel but this has not effect, I imagine it could also make the snapshot confusing. Is there a strong reason for why it is there?

qdev_wrappers/customised_instruments/parametric_sequencer.py
```diff
@@ -1,197 +1,459 @@
...
-                self.set_sqel_loopcnt_to_inf(i)
 
-    def _get_seq_mode(self):
-        if self.set_sqel_loopcnt() == '1':
-            awg_val = False
+        # add metadata, that gets added to the snapshot automatically
+        # add it before the upload so that if there is a crash, the
+        # state that is causing the crash is captured in the metadata
+        # TODO: add serialization of the elements
+        self.metadata['template_element'] = template_element
```
nataliejpg: maybe leave this commented until the serialisation is done so that we can in the meantime add the instrument to the station?

nataliejpg:
also can we maybe make this pr be just the sequencer and pull to the master branch so we start to move things onto there?

Dominik-Vogel:
> also can we maybe make this pr be just the sequencer and pull to the master branch so we start to move things onto there?

you mean without `testpulsbuilding.py`? that is a great idea, I'll pull it out.

Dominik-Vogel:
Thank you for your comments @nataliejpg ! 
- initialising without uploading seems like a good idea
- uploading of single element without setpoints also seems like a good idea, this is kind of the 0D case. I think we need to simply check if setpoints are None in some places and make sure we don't change the repeat mode to something illegal.
- duh, yeas, it is quite a bad idea to have mutable default arguments, thanks for catching that one.
- About the last point of where the parameters live, I am very interested in the experience you gathered and am excited to build a more usable solution with your input.

and oh yes absolutely shall we make this ultimately a pr against master.

Dominik-Vogel:
@ThorvaldLarsen Yes it would be desirable to have more than two dimensions, agreed. I decided to start implementing two dimensions to start with, as it is much easier and covers probably 90% of the use cases (higher dimensions will probably quickly make you run out of memory). When the 2d version is working it should be easy to generalise to n dimensions. Let me know about your specific needs and I will prioritize it accordingly.

PR #125, opened 2018-07-18
Alazar measuring only Channel A

ThorvaldLarsen:
There are issues with memory being cleared fast enough when we want to measure continuously. This PR allows to only measure channel A cutting the amount of data to be handled by 2.

This is really only a fix and should probably be done correctly at a later point (this will not work for channel B with the way the alazar_channels are set up). This is also not dynamic so if one changes the number of channels on alazar then one needs to manually update alazar channel controller accordingly.

@jenshnielsen


jenshnielsen:
With the caveats that you have mentioned I think this is ok to merge

ThorvaldLarsen:
There are quite a few things that might be good to go through with the alazar driver to increase the speed. We had a meeting with @Dominik-Vogel about this also, however it does not seem like there is manpower to do it atm. 
In my mind this can be added to the list of requested improvements.

PR #129, opened 2018-08-08
[WIP] add heterodyne source driver

nataliejpg:
represents two SGS100A instruments to set them a fixed 'demod_freqeuncy' apart for use with mixing up and down


nataliejpg:
@jenshnielsen this is tested

qdev_wrappers/customised_instruments/heterodyne_source.py
```diff
@@ -0,0 +1,128 @@
...
+        self._localos = localos
+        if cavity is None or localos is None:
+            raise RuntimeError('must initialise with microwave sources '
+                               'for cavity and local oscillator')
+        super().__init__(name)
+
+        self.add_parameter(name='frequency',
+                           set_cmd=self._set_drive_frequency,
+                           get_cmd=self._cavity.frequency,
+                           vals=vals.Numbers(1e6, 20e9))
```
jenshnielsen: I think it would be handy with a docstring stating explicitly how this sets the freqs of lo and cavity.
Something like:

docstring='Sets cavity frequency to the input frequency and LO frequency to input frequency + demodulation_frequency`

jenshnielsen: Same applies below

qdev_wrappers/customised_instruments/heterodyne_source.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def _get_ref_osc_external_freq(self):
+        cav_source_freq = self._cavity.ref_osc_external_freq()
+        lo_source_freq = self._localos.ref_osc_external_freq()
+        if cav_source_freq == lo_source_freq:
+            return cav_source_freq
+        else:
+            logging.warning('cavity and local oscillator do '
+                            'not have the same reference source frequency: '
+                            '{}, {}'.format(cav_source_freq, lo_source_freq))
```
jenshnielsen: As above

qdev_wrappers/customised_instruments/heterodyne_source.py
```diff
@@ -0,0 +1,128 @@
+from qcodes.instrument.base import Instrument
+from qcodes.utils import validators as vals
+import logging
+from qdev_wrappers.parameters import DelegateParameter
+
+# TODO: name help pleaaase!
+# TODO: should the cavity and local os actually be parameters?
```
jenshnielsen: Is is handy to be able to replace them? We probably want to make sure that it's captured in the snapshot which instruments they are

qdev_wrappers/customised_instruments/heterodyne_source.py
```diff
@@ -0,0 +1,128 @@
...
+        self._cavity.ref_osc_source(ref_osc_source)
+        self._localos.ref_osc_source(ref_osc_source)
+
+    def _get_ref_osc_source(self):
+        cav_source = self._cavity.ref_osc_source()
+        lo_source = self._localos.ref_osc_source()
+        if cav_source == lo_source:
+            return cav_source
+        else:
+            logging.warning(
```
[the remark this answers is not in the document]
nataliejpg: yes I think it should

qdev_wrappers/customised_instruments/heterodyne_source.py
nataliejpg: bad thing is that lots of the other parameters of the instrument are delegate parameters and so if it is possible to change the values of the InstrumentRefParameter then you could end up in a situation where InstrumentRefParameter points to another instrument but the delegate parameters are still on the original. It seems like overengineering to do a workaround for that.

nataliejpg:
- [ ] since update I need to test this and write an example

nataliejpg:
realistically might have been better to merge the original when I finished it and make changes as a serperate PR but we are where we are

PR #131, opened 2018-08-08
alazar_controller_setpoints

nataliejpg:
allows easier setting of setpoints names, labels, units using prepare_channel


nataliejpg:
@jenshnielsen I have by now tested most of the use cases of this, can we merge?

ThorvaldLarsen:
Reverting this commit as it breaks my code.
Please notify other users about merges before they happen.

@nataliejpg @jenshnielsen

Dominik-Vogel:
@ThorvaldLarsen you are right, we shouldn't cause people trouble by creating incompatibilities. To catch this more easily I will create a slack channel. Please help me to get everybody onboard of this channel.

PR #132, opened 2018-08-08
[WIP] add seq mode to awg customised instrument

nataliejpg:
TODO: 
- [x] TEST

### DIFF
--- a/qdev_wrappers/customised_instruments/AWG5014_ext.py
+++ b/qdev_wrappers/customised_instruments/AWG5014_ext.py
@@ -4,12 +4,40 @@ from qcodes.utils import validators as vals
 
 
 class AWG5014_ext(Tektronix_AWG5014):
-    def __init__(self, name, visa_address, **kwargs):
-        super().__init__(name, visa_address, **kwargs)
+    def __init__(self, name, address, **kwargs):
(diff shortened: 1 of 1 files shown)


nataliejpg:
@jenshnielsen tested

qdev_wrappers/customised_instruments/AWG5014_ext.py
```diff
@@ -4,12 +4,39 @@
...
+            for i in range(self.sequence_length()):
+                self.set_sqel_loopcnt_to_inf(i)
+
+    def _get_seq_mode(self):
+        if self.get_sqel_loopcnt() == '1':
+            return False
+        else:
+            return True
+
+    def send_and_load_awg_file(self, awg_file, filename):
```
jenshnielsen: This function looks general enough that it perhaps belongs in the driver?

nataliejpg: Possibly. I think up to taste to be honest, 'make' is already in there as are 'send' and 'load' and a few combinations. Maybe @WilliamHPNielsen has an opinion?

nataliejpg:
No longer necessary as we expect to use the awg with the parameteric sequencer and so this is overruled by the repest_mode parameter on that

PR #133, opened 2018-08-08
alazar_seq_mode_to_bool

nataliejpg:
WYSIWYG

TODO: 
- [x] test

### DIFF
--- a/qdev_wrappers/customised_instruments/AlazarTech_ATS9360_ext.py
+++ b/qdev_wrappers/customised_instruments/AlazarTech_ATS9360_ext.py
@@ -8,9 +8,7 @@ class AlazarTech_ATS9360_ext(AlazarTech_ATS9360):
         self.add_parameter(name='seq_mode',
                            get_cmd=self._get_seq_mod,
                            set_cmd=self._set_seq_mode,
-                           vals=vals.Enum('on', 'off'),
-                           get_cmd=get_parser_on_off,
-                           set_cmd=set_parser_on_off)
+                           vals=vals.Enum('on', 'off', True, False, 1, 0))
 
     def _get_seq_mod(self):
         if (self.aux_io_mode() == 'AUX_IN_TRIGGER_ENABLE' and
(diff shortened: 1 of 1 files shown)


qdev_wrappers/customised_instruments/AlazarTech_ATS9360_ext.py
```diff
@@ -8,24 +8,30 @@ def __init__(self, name):
         self.add_parameter(name='seq_mode',
                            get_cmd=self._get_seq_mod,
                            set_cmd=self._set_seq_mode,
-                           vals=vals.Enum('on', 'off'))
+                           vals=vals.Enum('on', 'off'),
+                           get_cmd=get_parser_on_off,
+                           set_cmd=set_parser_on_off)
```
jenshnielsen: I can't seem to spot that this is imported anywhere?

nataliejpg: you're right, let me fix that

nataliejpg:
have tested and it works, I realise this is slightly api changing (in terms of returning a bool instead of a string) but if I make a shout out on the qcodes announcements slack can I merge this?

jenshnielsen:
Yes that's probably fine. @ThorvaldLarsen?

ThorvaldLarsen:
Looks good to me. 

@jenshnielsen is there any chance that there will be a qcodes parser for Boolean parameters? Almost all instruments has at least one such parameter (e.g. output on/off) and would be nice to have a common behavior for all instruments.

ThorvaldLarsen:
Yes, guess I knew that there was a issue with it... 
However, looking at it seems everyone agrees and it is just a question of where the PR goes to main qcodes. It could live next to validators but maybe as a separate file in utils.

nataliejpg:
I'll merge this now, @ThorvaldLarsen and @jenshnielsen maybe continue the on/off discussion in the issue I mentioned?

jenshnielsen:
Yes I think we should just bring this thing into qcodes as is

PR #134, opened 2018-08-20
update decadac customised instrument

nataliejpg:
so that it uses the station configurator and yaml rather than the old config file version


jenshnielsen:
👍  Do you think anyone is using the old version?

nataliejpg:
I don't think so. I'll ask on slack

nataliejpg:
I couldn't find anyone on the 4th floor using the decadac with qcodes and the old config file the two times I went looking. Can we merge this?

PR #135, opened 2018-08-22
Smart color scale

Dominik-Vogel:
This PR adds an automatic color scaling for saved heatmap images (using doNd). All the 1% of the lowest and highest values will be represented in the same color (each). This means that outliers as long as they make up less than 1% of the data will be suppressed.

This is meant to bridge the gap until the new data set is fully rolled out in Copenhagen and gather some experience with the color scaling.

### DIFF
--- a/qdev_wrappers/plot_functions.py
+++ b/qdev_wrappers/plot_functions.py
@@ -1,3 +1,4 @@
+from typing import Tuple
 from os.path import sep
 from copy import deepcopy
 import functools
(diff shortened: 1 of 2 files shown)


Dominik-Vogel:
@nataliejpg  and @ThorvaldLarsen it would be great if you could give quick feedback so that we can merge it soon and have some time before we implement for the new dataset.

nataliejpg:
@Dominik-Vogel I'm using plot_by_id by now, I thought that this didn't use qc.MatPlot etc? If it does not then I won't be able to try it out. Otherwise looks good to me.

ThorvaldLarsen:
This is a great idea! Would definitely be very usefull. Two comments:
- would be good to have the 1 % threshold defined somewhere that is changeable rather than hardcoded (qcodes config file or yaml?). I cannot currently comment on what a meaningful value would be which could also very likely depend on what experiment is being run.
- Follow up on Natalie's comment: as we slowly prepare to move to new dataset we should make sure that all the functionalities gets ported to plot_by_id/QCplot (or whichever plot function we should be using).

Dominik-Vogel:
ok thanks for the feedback. Good to see that you are already using plot_by_id.  I also thought about having margins configurable...

jenshnielsen:
I agree on making it configurable but something like https://www.khanacademy.org/math/statistics-probability/summarizing-quantitative-data/box-whisker-plots/a/identifying-outliers-iqr-rule may be more correct than a fixed percentage but we need to evaluate the performance cost of calculating the IQR

jenshnielsen:
It may also be useful to use the colorbar extend as documented here https://matplotlib.org/api/_as_gen/matplotlib.pyplot.colorbar.html

ThorvaldLarsen:
Actually on second thought it might be fine to have it somewhat hardcoded as long as there is an easy way to manually specify range for the colorbar.

Concerning the implementation I dont think anyone is using `_plot_setup` with `useQT=False`. This function is only used for the Liveplotting where everyone is using QTplot to my knowledge. If we want to implement it for old dataset it should go in `show_num` - but might be better to just focus on `plot_by_id`.
It might also be worthwhile to make sure this works for plots combined from multiple dataset. That is the percentages needs to take all dataset into account for the calculation. And depending on the implementation each dataset might have separate colorbars in matplotlib.

Dominik-Vogel:
Thank you all for the feedback, that is what I was hoping for. I am looking into IQR and the colorbar extend feature. I will port things over to `plot_by_id`.

(@ThorvaldLarsen _plot_setup is actually used when saving images with matplotlib, this is how the ones I posted have been created)

ThorvaldLarsen:
@Dominik-Vogel I see - forgot about that. However, we will probably not be able to help out much in the V0 team to test this as we almost always need the functionalities in `show_num` to view the data in a useful fashion.

PR #137, opened 2018-08-31
Fix customised instrument args

nataliejpg:
ZNB_ext and AWG5014_ext to have address instead of visa_address as init args so that they plan nice with the station configurator.


jenshnielsen:
👍  I guess this could break for some users if they are using visa_address as a kwarg but it's probably not worth the effort to make clever workarounds

PR #138, opened 2018-09-03
Updating comments on installation script

Dominik-Vogel:
WYSIWYG

### DIFF
--- a/installation/install.ps1
+++ b/installation/install.ps1
@@ -1,4 +1,9 @@
-﻿# This script installs all you need to use qcodes:
+# To run this installation script, right-click it and select run with PowerShell
+# To run it from ISE execute:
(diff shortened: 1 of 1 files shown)


installation/install.ps1
```diff
@@ -1,4 +1,9 @@
-﻿# This script installs all you need to use qcodes:
+# To run this installation script, right-click it and select run with PowerShell
+# To run it from ISE execute:
+# `Set-ExecutionPolicy -Scope Process -ExecutionPolicy Unrestricted`
+# to temporarily elevate the rights to executed foreign scripts
+
+# This script installs all you need to use qcodes:
```
jenshnielsen: This is not completely true is it :) We are missing NI Visa?

Dominik-Vogel: good point

PR #139, opened 2018-09-03
Parametric sequencer

Dominik-Vogel:
(Carrying over the parmetric sequencer from the cavity branch)
The parametric sequencer allows the user to comfortably control AWGs to execute sequences and provides qcodes parameters for the doNd wrappers.

needs Flatbean version of Broadbean to work (https://github.com/QCoDeS/broadbean/pull/90)
currently only the AWG5014 is supported. Extensions to other AWGs can be written easily


nataliejpg:
One thing that victoria and I noticed was that with repeat mode 'element' when setting the values of repeat parameters they were not changed immediately on the instrument so that if you did a sweep over one of their values it wouldn't work as you end up measuring while the instrument is setting the element. Adding a time.sleep fixed this but I imagine that you might want something more robust. Just a heads up

nataliejpg:
Also porting over my comments from #124, maybe they should go into an issue if we want to merge this branch and can be part of the 'second generation' but in the meantime I didn't want them to get lost.
-  would be great if you could initialise without having to upload a sequence already
-  would be great if you could also have the option to just upload one element and not have any setpoints
-  could you add a warning if the user tries to set the value of a sequence parameter which is already being varied across the sequence. I understand the need to have the repeat parameters also on the sequence channel but I think that it is problematic if you can set them without being warned especially because at the moment setting them has no effect.

nataliejpg:
- [ ] if you specify a channel in the yaml but it does not match a real channel name or a routed channel name it would be nice to have at least a warning that this isn't being uploaded and maybe you've made a typo

Dominik-Vogel:
There are two open points left. The first one, I didn't get to, but it should be easy to implement. The second one is trickier, because it is basically a hardware limitation of the AWG5014C (as we talked about).

PR #142, opened 2018-09-09
Revert "alazar_controller_setpoints"

ThorvaldLarsen:
Reverts qdev-dk/qdev-wrappers#131


nataliejpg:
@ThorvaldLarsen can you tell me what was actually broken that you needed to roll back #131? I've had a look and can't see how it should change the default behaviour so if you have a link to something that actually breaks or some time to go through it that would be great.

ThorvaldLarsen:
I didn't look too closely at it but I think it was related to chan.data.setpoints being empty just after initialization. Can you maybe make a new PR with the changes (with pep 8 separated)? Then I can also take a look.

PR #143, opened 2018-09-20
Add data exporter

nataliejpg:
adds function which exports a run from the database to a .txt file saved in the same folder as the live-plotting pngs are saved. I imagine that there will at some point be something in qcodes for this but in the meantime people might want to use it :)


Dominik-Vogel:
Let me write some tests for it and put it in qcodes as soon as possible. Nevertheless I will merge this already to keep the spirit of the wrappers.

PR #144, opened 2018-09-21
updates setting setpoints on alazar channel data parameter

nataliejpg:
setting setpoints can now be done via set_setpoints_and_labels function which is also called by prepare_channel. The setpoints, names, labels and units. See example notebook for details. Default behaviour remains the same

- [x] test


ThorvaldLarsen:
Anders has been running with this PR for a week without any issues. So this has been tested with the cQED code at T3,T11,QT2.

nataliejpg:
great, thanks @ThorvaldLarsen. We've been using it at T2 as well on both measurement pcs but I didn't test the two dimensional parameters yet and was thinking I should run the example notebook to check that works too. I will try to do that today or early next week though and then we can merge

nataliejpg:
ok this is now ready I think. I've been using it for 1d traces with demodulation for a few months and have tested the example notebook for 1d and 2d traces (but without only noise input) @ThorvaldLarsen or @jenshnielsen could you review?

nataliejpg:
also I couldn't resist do I deleted the "Qcodes example with Alazar ATS9360 old controller" notebook and the "acquisition_parametersold.py" because unless I'm wrong(?) its pretty much impossible that anyone is using them @jenshnielsen

PR #147, opened 2018-10-05
Revert "Alazar measuring only Channel A"

ThorvaldLarsen:
@jenshnielsen I have forgotten to revert this. Had some troubles with this PR as the testing I did was not performed on an actual sample so only tested that the alazar returns data.
After running on a sample it seems that it somehow changes the order of records in the returned data matrix (or the demodulation code somewhere assumes channel A and B records are interleaved).

I propose reverting this and then returning to the issue of single channel measurements at a later point.

Reverts qdev-dk/qdev-wrappers#125 @nataliejpg


jenshnielsen:
Ok to merge this from me

PR #153, opened 2018-11-21
add: T12 stuff

MerlinSmiles:
adding stuff from T12 setup


Dominik-Vogel:
@MerlinSmiles  could I ask you to extract the parts that are potentially of general interest into a top level folder. We decided that we would try to avoid having fridge specific code as people and experiments kept moving around and useful code spread across many stations. (This might not be the case for everything you commited but from a brief glance it seems that there are some things that might be of general interest).

qdev_wrappers/T12/merlin/timeinstrument.py
```diff
@@ -0,0 +1,23 @@
...
+from qcodes import Instrument
+
+
+class TimeInstrument(Instrument):
+    def __init__(self, name, **kwargs):
+        super().__init__(name, **kwargs)
+
+        self.add_parameter('time',
+                           unit='s',
+                           get_cmd=time.time,
```
Dominik-Vogel: Just a cautious warning: we have been experiencing problems with `time.time` as it does not guarantee you any precision, it turns out that when calling it consecutively you don't necessarily get a monotonous result. As an alternative there is the perf_counter.

Dominik-Vogel:
What is the zip file about?

MerlinSmiles:
> 
> 
> What is the zip file about?
ups, i think that was when william updated qcodes with a plan to put all this in the wrappers

MerlinSmiles:
@Dominik-Vogel 
not sure when I'll have the time/ambition to make things a grenerally beauty to add this to something top-level, i guess i might keep it on my github then ...

Dominik-Vogel:
I am not having high quality requirements. I was more thinking about simply moving files.... I didn't mean to cause you a lot of work for that. I think it is great if we all can profit from a common code base.

MerlinSmiles:
sure, it wouldnt require much time, just figuring out what all that majorana spinqubit and whatnot stuff is about and where these things would go. Just wont do that at the moment :)

MerlinSmiles:
oh and thanks for the time comment, i totally forgot about that perf_counter

Dominik-Vogel:
Ah, I see, these will be deleted once we made sure nobody is using them anymore. Instruments generally just go into the customised instruments folder. If you have a new category feel free to create a folder for that.

PR #156, opened 2018-12-05
Refactor customised instruments

### LINKED ISSUE #154: restructure
How do people feel about restructuring the repository? I know that this by default is pretty annoying but I think it would be a lot easier to use and save a lot of headache later. My ideas are

-  subfolders in the customised_instruments folder by name of the instrument manufacturer or user defined name, this would mirror the qcodes structure and is useful for cases of multiple customised channels which belong to one instrument or similar.
- a folder for customised_parameters where customised parameters live, the argument against this is that they might be very specific and only work with a particular (customised) instruments and might be better grouped with them
- put the alazar_controller folder inside customised instruments (and pull out the parameters to live in customised_parameters if we go ahead with that). 
- delete the subgroup folders (transmon, majorana, spinqubit). If anyone is still using anything from them then it makes more sense to refactor it into somewhere existing or think of a new place to put it named based on functionality rather than the name of the subgroup expected to use it...

@jenshnielsen @WilliamHPNielsen

nataliejpg:
- Moves customised instruments into folders named after the make. This includes putting the alazar_controllers folder into customised_instruments.
- Deletes the subgroup folders (since this code should be local or refactored into somewhere else in the repository)
- moves export function into the dataset folder

fixes #154

@ThorvaldLarsen @jenshnielsen @Dominik-Vogel @fabioansaloni


nataliejpg:
also fulfills #56

PR #157, opened 2018-12-05
plot saving to be specified in config.user

nataliejpg:
When using doNd if the 'do_plots' keyword argument is True plots will be generated which are saved in both png and pdf format. This pull request makes the saving configurable via the config file so that you can choose if you want any generated plots to be saved as png, pdf, both or neither :)

TODO:
- [ ] Test

@jenshnielsen


nataliejpg:
NB keys in config are now "save_pdf" and "save_png".

nataliejpg:
I've updated this and tested it so that it definitely works but I realise it conflicts with #166. Since this version is really not a big change I'd be happy for that pr to get merged first and then I update this one but I'd really like it to be soon (like in the next week kind of soon) or to get this one merged instead @ThorvaldLarsen @jenshnielsen thoughts?

PR #159, opened 2018-12-06
[WIP] create interface parameter

nataliejpg:
Adds a parameter to the customised instruments which can function either as a delegate parameter, or be given function to run for get and set (minimally these are none in which case it behaves like a manual parameter). The thinking behind this is that it makes it much easier to create instrument interfaces with designated parameters and then subclasses can then attach these parameters to functions or real instruments but the higher level code is minimally affected by changing physical instruments and experiment code can be easily tested using simulated instruments.

- [ ] test


nataliejpg:
replaced by #165

PR #166, opened 2018-12-28
Changes to doNd and plot_by_id

ThorvaldLarsen:
Making a few changes for the new dataset functions in qdev_wrappers.

Changes in this PR:

- Modify doNd functions to allow arrayparameters of different lengths.
- Save arrayparameters as blobs rather than individual points
- Change save_image to take data_id as input to allow use outside doNd functions
- Move save_image to plotting file
- Copy plot_by_id from main qcodes to add functionality to subtract averages from rows/columns and transpose plotted data (This should not be standard but think it is okay short term instead of modifying plot_by_id in main qcodes).

@WilliamHPNielsen @jenshnielsen @nataliejpg


ThorvaldLarsen:
@WilliamHPNielsen I added the transformation functions here. Also removed the copy of plot_by_id.

nataliejpg:
is there an example notebook with any of these things in? I think that would be really valuable in general but especially in moments like this it would make it really quick to see the functionality. Otherwise sounds really good and we are definitely very keen for the row/column average tool!

ThorvaldLarsen:
@nataliejpg I have added an example notebook for plotting. There is already one for doNd which I think is still up to date - this PR only fixes some bugs in doNd not their functionality.

ThorvaldLarsen:
@jenshnielsen when you get a chance, can you take a quick look at this and maybe merge it? Only thing not working is multiple arrayparameters of different size, which I believe is also not working without this update to doNd functions.

nataliejpg:
i like it a lot. I would add the gluing datasets together as something desirable but also see no reason not to go ahead with this merge in the meantime.

ThorvaldLarsen:
@jenshnielsen this PR has been tested quite a bit at T3. Any chance we can merge it?

jenshnielsen:
@ThorvaldLarsen  I pushed up the changes to unify the progessbar with the rest of qcodes. I think this is ready to merge

PR #167, opened 2018-12-28
Remove logger

ThorvaldLarsen:
With logger now in main qcodes i think we should remove it from qdev_wrappers.

@Dominik-Vogel


WilliamHPNielsen:
@ThorvaldLarsen, rather than deleting the `qdev-wrappers/logger.py`, I think it's a bit more human-friendly  to put in a deprecation warning saying that this file is outdated and that the qcodes one should be used instead.

ThorvaldLarsen:
This is probably a much better way to do this. Would you be able to add the warnings (maybe in a new PR if easier)? Not really comfortable with the python warnings...

WilliamHPNielsen:
@ThorvaldLarsen, sure thing. I'll do that now.

WilliamHPNielsen:
Closing this in favour of #168

PR #173, opened 2019-01-23
Added conductance buffer parameter and parameters for current biased …

ectof:
…differential resistance measurement

Conductance buffer class existed previously but wasn't added as a parameter, this seems to have been a bug. Also added some new parameters for current biased resistance measurements


ectof:
Yes, I tried these out on T11, but I realized that the label is wrong for the resistance buffer parameter, it's now labelled conductance, should be "Resistance", maybe you can fix it on master when you have a chance

PR #174, opened 2019-01-25
newer fitter for wrappers

lillian542:
In progress extension to fitting compatible with new dataset.
includes
- [ ]   fitter (based on @ThorvaldLarsen qdev_fitter.py). Uses a 'model' to perform a fit
- [ ]   fitter.fit - takes arrays, and returns a dictionary of fit parameters
- [ ]   fitter.fit_by_id - takes data run id from saved data (and dataname information) and returns a 'fit'
- [ ]   fit ('analysis') object returned by the fitter
- [ ]   plotting of results for fits returned by the fitter (and doesn't use any qcodes or qdev wrappers)
- [ ]   save function to store resulting fit as its own dataset with a separate run id in the database
- [ ]   plotter which can plot data from a dataset created by the step above

All current models are for fitting 1d functions to 1d data, or 1d cuts of Nd data.


lillian542:
Could you review this for merging? @WilliamHPNielsen @jenshnielsen @Dominik-Vogel @astafan8

nataliejpg:
there's a stray jupyter notebook called Untitled.ipynb floating around on the top level and if the examples could be in the examples folder rather than in the fitting that would be great.

nataliejpg:
Ok it's 'done' at least for the first round. @Dominik-Vogel @jenshnielsen @WilliamHPNielsen @ThorvaldLarsen could some of you take a look? The only thing blocking right now is that it relies on #157 (just because I wanted to use the same logic for saving plots we generated so had to pull plot_by_id out of save_image) and that in turn conflicts with #166 BUT it's still very much ready for people to look at and if we could get #166 merged asap I can update #157 and then this is good to merge and I think it's really due. I would also argue for deleting the other qdev_fitter but only once this is merged and people using the old one have moved over.

nataliejpg:
the other thing is that I added a database to the examples because you need data to fit to and it will also come in handy for the optimisation examples we'll want to add really soon. It's only 10MB (8 datasets, the pngs are in the same folder) but if that seems like it's going to be a problem I can either slim it down or delete it and rely on users having their own data to fit to for the fitter examples or make the example notebooks read only. If we do leave the database there it seems reasonable to add it to the gitignore or by some other method make it harder to push changes so that running the example notebooks doesn't mean you also have to remember to reset you database afterwards.

PR #181, opened 2019-04-12
Remove logging warnings 

jenshnielsen:
Don't have file setup import deprecated module but call the new logging module in qcodes. 

Remove import of logger in module `__init__` file

@Dominik-Vogel  can you have a look


WilliamHPNielsen:
I like the changes introduced here. While we're at it, shouldn't we fix those deprecation warnings that point to deprecated code? Perhaps they have outlive their cycle now and can be removed? I'm thinking about the warnings in `init_python_logger` and `_set_up_ipython_logging`.

jenshnielsen:
@WilliamHPNielsen  Updated the warnings

PR #186, opened 2019-06-07
Add `write_period` as a kwarg to `dond`s

Dominik-Vogel:
Currently there is no way of setting the `write_periode` in `dond` calls. The implementation of this keywords follows the wish to have faster live plotting.

### DIFF
--- a/qdev_wrappers/dataset/doNd.py
+++ b/qdev_wrappers/dataset/doNd.py
@@ -20,6 +20,7 @@ number = Union[float, int]
 
 
 def do0d(*param_meas:  Union[_BaseParameter, Callable[[], None]],
+         write_period: Optional[float] = None,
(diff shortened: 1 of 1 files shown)


qdev_wrappers/dataset/doNd.py
```diff
@@ -37,6 +38,8 @@ def do0d(*param_meas:  Union[_BaseParameter, Callable[[], None]],
         The run_id of the DataSet created
     """
     meas = Measurement()
+    if write_period:
```
WilliamHPNielsen: I guess this will automatically protect the user from setting the write period to `0.0`, but perhaps in a slightly to "clever" way. Perhaps `if write_period is None` and an explicit handling of 0 is better?

PR #193, opened 2019-11-12
Refactor doNds

Dominik-Vogel:
Lets create some tests to see if I broke anything ;-)


Dominik-Vogel:
@GateBuilder would you be interested in writing the tests?

Dominik-Vogel:
Here is already one I had used in another context:
```python
def test_do0d():
    from qdev_wrappers.dataset.doNd import do0d, do1d
    from qcodes.instrument.parameter import Parameter
    from qcodes import config
    from qcodes.utils import validators
    config.user.mainfolder = "output"  # set ouput folder for doNd's

    s = Parameter('setter', set_cmd=None, get_cmd=None)
    p = Parameter(
        'test',
        set_cmd=None,
        get_cmd=lambda: 1+1j,
        vals=validators.ComplexNumbers())
    do1d(s, 0, 1, 1, 0, p)
    do0d(p)
```

qdev_wrappers/dataset/doNd.py
```diff
@@ -1,27 +1,90 @@
...
+
+
+def _register_parameters(
+        meas: Measurement,
+        param_meas: List[ParamMeasT],
+        setpoints: Optional[List[_BaseParameter]] = None
+) -> None:
+    for parameter in param_meas:
+        if isinstance(parameter, _BaseParameter):
+            meas.register_parameter(parameter,
```
ThorvaldLarsen: @Dominik-Vogel is it possible that we can handle array parameters separately to ensure they get saved as blobs in the database. This was the main reason this PR https://github.com/qdev-dk/qdev-wrappers/pull/166 exists.

As is the doNd functions are completely unusable when doing anything with an arrayparameter (the overhead for saving is way too large).

Dominik-Vogel: Good point! Thank you for notifying us!

Dominik-Vogel: We will consider these in  a new pr, to keep this one pure refactoring.

Dominik-Vogel: @GateBuilder can you work on a test for this?

PR #195, opened 2019-11-14
Implement tests for doNd

GateBuilder:
This PR introduces basic black box tests for doNd function of Wrappers.

@Dominik-Vogel


qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,92 @@
...
+
+_paramComplex = Parameter('simple_complex_parameter',
+                   set_cmd=None,
+                   get_cmd=lambda: 1 + 1j,
+                   vals=validators.ComplexNumbers())
+
+_param_set = Parameter('simple_setter_parameter',
+                       set_cmd=None,
+                       get_cmd=None)
+
```
Dominik-Vogel: These should be fixtures to make sure that we are independent of the state of the parameters.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _param_func():
+    """
+    A private utility function.
+    """
+    _new_param = Parameter('modified_parameter',
+                           set_cmd= None,
+                           get_cmd= lambda: _param.get()*2)
+    assert _new_param.get() == 2
```
Dominik-Vogel: We might be able to make this a little cleaner by separating clearly between tests and fixtures. This function is both (uses assert and returns something that gets used somewhere else)

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,92 @@
...
+    return _new_param
+
+
+def test_do0d():
```
Dominik-Vogel: You can shorten this test by parametrizing it.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_do1d():
+
+    _start = 0
+    _stop = 1
+    _num_points = 1
+    _delay_list = [0, 0.1, 1]
+
+    for _delay in _delay_list:
```
Dominik-Vogel: delay list could also be a parameter

qdev_wrappers/tests/dataset/test_doNd.py
GateBuilder: Does parametrization has a performance advantage? If not, I prefer to have every statement in the test case explicitly written.

qdev_wrappers/tests/dataset/test_doNd.py
GateBuilder: Sure, but this is an optional argument of the function being tested. Parameters are fixtures which are not the subject of the test, while _delay_list is. So it should stay as it is, IMHO.

qdev_wrappers/tests/dataset/test_doNd.py
GateBuilder: Making parameters fixtures sufficient for this distinction? Or you would like to see something here explicitly?

Dominik-Vogel: Making them fixtures is a good thing, and then adding a simple test that uses that fixture and does the assert, instead of having the assert inside the fixture.

qdev_wrappers/tests/dataset/test_doNd.py
Dominik-Vogel: as per offline discussion: parameter in the sense of pytest.parametrize not qcodes.Parameter.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def test_param_callable(_param_callable):
+    _param_modified = _param_callable
+    assert _param_modified.get() == 2
+
+
+@pytest.mark.parametrize('period, plot', [(None, True), (None, False),
+                         (1, True), (1, False)])
+def test_do0d(_parameters, _param_callable, period, plot):
```
Dominik-Vogel: This could be split up into individual tests, for the different scenarios and it could be fused with the other tests by adding more asserts:
so e.g. test do0d with a simple parameter and assert if the output is correct.
Another test could be calling a do0d with a callable and a parameter and check if they are handled in the right order.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,165 @@
...
+    # satisfied.
+
+    do0d(_param, write_period=period, do_plot=plot)
+    do0d(_paramComplex, write_period=period, do_plot=plot)
+    do0d(_param_callable, write_period=period, do_plot=plot)
+    do0d(_param, _paramComplex, write_period=period, do_plot=plot)
+    do0d(_param_callable, _paramComplex, write_period=period, do_plot=plot)
+
+
+def test_do0d_out_type_1(_parameters):
```
Dominik-Vogel: Ideally the names of the tests reveal what they are doing.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -163,3 +189,31 @@ def test_do2d_out_type(_parameters):
                  _param, _paramComplex)
 
     assert type(_data[0]) == int
+
+
+def test_do2d_data(_parameters):
+
+    _start_p1 = 0
```
Dominik-Vogel: all identifiers that are in a private scope anyways don't need an additional `_` . Here e.g. `_start_p1` cannot be accessed from any other place than from within the test function.
For module level functions, like the fixtures you are defining, I feel that you are right if you are very strict about it, but it would also be ok to not have the `_` in the fixtures.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,165 @@
...
+    return _new_param
+
+
+@pytest.fixture()
+def _param_callable(_parameters):
+    _param, _, _ = _parameters
+    return _param_func(_param)
+
+
+def test_param_callable(_param_callable):
```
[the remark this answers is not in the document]
GateBuilder: This is because I inserted the "assert" within the function itself just for sanity check in the first version. This check does not require a separate test and even can be completely omitted. I prefer to loose the test and move assert into the fixture.

qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -0,0 +1,165 @@
...
+    A private utility function.
+    """
+    _new_param = Parameter('modified_parameter',
+                           set_cmd= None,
+                           get_cmd= lambda: _p.get()*2)
+    return _new_param
+
+
+@pytest.fixture()
+def _param_callable(_parameters):
```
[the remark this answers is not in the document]
GateBuilder: I am confused by this comment. I do implement this test with an arbitrary callable, but just to be fair to the context I created a function which manipulates a QCoDeS parameter; not because I think Callables are related to QCoDeS parameters :)  

The main purpose here to test if doNd work with an arbitrary function as its positional argument. That is it. 

I am also missing the point of your example; do we really need to create a callable object?

qdev_wrappers/tests/dataset/test_doNd.py
GateBuilder: I would prefer the keep the tests for run-time and return-values and -types seperately. Thus, as I learnt from our offline disscussions, because parametrize creates these seperate tests internally, I would keep the current implementation.

Dominik-Vogel: Ok, fair enough, that addresses the point of fusing some tests, but how about splitting and asserting order?

qdev_wrappers/tests/dataset/test_doNd.py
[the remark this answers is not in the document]
Dominik-Vogel: Fixtures should ideally not do assertions.

qdev_wrappers/tests/dataset/test_doNd.py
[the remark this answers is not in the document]
Dominik-Vogel: Ok, great, I simply misunderstood you there, because going through a qcodes parameter does not seem like the most obvious solution to me. I was wondering why you did not simply do this:
```python
@pytest.fixture
def callable():
    return lambda: 1
```
This does exactly the same as currently.

Concerning my sample code in my previous comment:
I created a callable class so that we can attach a state that we can readout, here the time stamps when it has been called. This could then help us to write a test which verifies the order in which callables and parameters have been called.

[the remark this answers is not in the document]
GateBuilder: I have decided the let this function and the corresponding test stays as it is; but they can later be improved on QCoDeS site, if needed.

qdev_wrappers/tests/dataset/test_doNd.py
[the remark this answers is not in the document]
GateBuilder: See my answer above.

PR #196, opened 2019-11-28
Add support for ArrayParameter in DoNds

Dominik-Vogel:
This is inspired by #166

see diff here:
https://github.com/qdev-dk/qdev-wrappers/compare/80516e7660355501cbd8f9545316f1a374298af0...ThorvaldLarsen:doNd_save_image

### DIFF
--- a/qdev_wrappers/dataset/doNd.py
+++ b/qdev_wrappers/dataset/doNd.py
@@ -7,8 +7,9 @@ import numpy as np
 import matplotlib
 import matplotlib.pyplot as plt
 
+from qcodes.instrument.parameter import (
+    _BaseParameter, ArrayParameter, MultiParameter)
 from qcodes.dataset.measurements import Measurement, res_type, DataSaver
-from qcodes.instrument.base import _BaseParameter
 from qcodes.dataset.plotting import plot_by_id
 from qcodes import config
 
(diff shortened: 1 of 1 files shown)


Dominik-Vogel:
@ThorvaldLarsen what is this sleep about?
https://github.com/ThorvaldLarsen/qdev-wrappers/blob/9df2e3359804a9dc4e9172a7de4194abb25e8080/qdev_wrappers/dataset/doNd.py#L112

Dominik-Vogel:
I have looked at the test and realized, that it is not actually necessary. Maybe it was when the changes first were made.
@ThorvaldLarsen 
https://github.com/QCoDeS/Qcodes/blob/bc0118c1ee4b5a4f54b2a56bf2dbe1adcde4b59f/qcodes/dataset/measurements.py#L869-L896

jenshnielsen:
The multiparameter part is afaik still not handled within qcodes

Dominik-Vogel:
True, but this we should handle in qcodes.

ThorvaldLarsen:
Not sure about the sleep. Also this PR is quite old so seems very plausible that what it solved has since been fixed in qcodes.

ThorvaldLarsen:
Well not this PR, but the original one I made to handle array paramters with new dataset

PR #197, opened 2019-11-28
DoNd register parameters

emartinez:
This pull request adds a register_params argument to doNd functions, to register in the measurement additional setpoint parameters that are not being changed. This is useful e.g. when a N-dimensional scan is broken over several slices, to facilitate reconstructing the entire N-dimensional measurement.

### DIFF
--- a/qdev_wrappers/dataset/doNd.py
+++ b/qdev_wrappers/dataset/doNd.py
@@ -248,7 +248,7 @@ def do2d(
 
                     datasaver.add_result((param_set1, set_point1),
                                          (param_set2, set_point2),
-                                         *_process_params_meas(param_meas))
+                                         *_process_params_meas(params_to_measure))
                 for action in after_inner_actions:
                     action()
                 if flush_columns:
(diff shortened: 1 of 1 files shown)


qdev_wrappers/dataset/doNd.py
```diff
@@ -141,15 +142,21 @@ def do1d(
...
     """
     meas = Measurement()
-    _register_parameters(meas, (param_set,))
-    _register_parameters(meas, param_meas, setpoints=(param_set,))
+    if register_params is None:
+        register_params = tuple()
+    setpoint_params = (param_set,) + tuple(register_params)
+    _register_parameters(meas, setpoint_params)
+    _register_parameters(meas, param_meas, setpoints=setpoint_params)
+    params_to_measure = param_meas + tuple(register_params)
```
GateBuilder: @emartinez Could you elaborate the reasoning to add `register_params` as both set points and also parameters to be measured? Are there any down falls of this approach? Do we have any performance degradation?

PR #198, opened 2019-12-11
Remove _latest from alazar controller

Dominik-Vogel:
With the update of the cache the private property `_latest` of the `Parameter` is no longer available. This fixes the usage in the Alazar controller. @astafan8

(The save_val I removed is not a bug, but as it will be deprecated I removed it in one go. The cache will be updated anyways in the set wrapper, so it does not need to be called)

### DIFF
--- a/qdev_wrappers/alazar_controllers/acquisition_parameters.py
+++ b/qdev_wrappers/alazar_controllers/acquisition_parameters.py
@@ -37,10 +37,9 @@ class AcqVariablesParam(Parameter):
             value: value to set the parameter to
         """
         self._check_and_update_instr(value, param_name=self.name)
-        self._save_val(value)
 
     def get_raw(self):
-        return self._latest['value']
+        return self.cache.get()
 
     def to_default(self):
         """
@@ -63,7 +62,7 @@ class AcqVariablesParam(Parameter):
         Return:
             True (if no errors raised when check_and_update_fn executed)
         """
-        val = self._latest['value']
+        val = self.cache.get()
         self._check_and_update_instr(val, param_name=self.name)
         return True
 


jenshnielsen:
Should we put a minimum qcodes version in the setup.py file?

qdev_wrappers/alazar_controllers/acquisition_parameters.py
```diff
@@ -37,10 +37,9 @@ def set_raw(self, value):
             value: value to set the parameter to
         """
         self._check_and_update_instr(value, param_name=self.name)
-        self._save_val(value)
 
     def get_raw(self):
-        return self._latest['value']
+        return self.cache.get()
```
astafan8: instead of this, just `pass get_cmd=None` to the constructor. This way the change will be qcodes backwards compatible.

qdev_wrappers/alazar_controllers/acquisition_parameters.py
```diff
@@ -63,7 +62,7 @@ def check(self):
         Return:
             True (if no errors raised when check_and_update_fn executed)
         """
-        val = self._latest['value']
+        val = self.cache.get()
```
astafan8: in case these parameters are (or happen to be) always such that they `value`s are always equal to their `raw_value`s, then this can be implemented in a backwards compatible manner via `val = self.raw_value`. But if not, then i'd rethink the purpose of this method - as far as i understood it, it basically does the same as `set_raw` but it takes the value to validate from "cache" as opposed to "from outside" - i wonder what the benefit of this functionality is, because thanks to `set_raw` the value in "cache" can't be not-valid, right?

qdev_wrappers/alazar_controllers/acquisition_parameters.py
```diff
@@ -37,10 +37,9 @@ def set_raw(self, value):
             value: value to set the parameter to
```
astafan8: this PR should also include a fix for [this `_save_val` call](https://github.com/qdev-dk/qdev-wrappers/blob/3412bc01e13060ea88428a876477700d9e0e5efb/qdev_wrappers/alazar_controllers/acquisition_parameters.py#L129). So far as i understand it, that call is just not needed (and was never needed in fact)

qdev_wrappers/alazar_controllers/acquisition_parameters.py
```diff
@@ -37,10 +37,9 @@ def set_raw(self, value):
...
         """
```
astafan8: not sure it's relevant for this PR, but [`NonSettableDerivedParameter`](https://github.com/qdev-dk/qdev-wrappers/blob/3412bc01e13060ea88428a876477700d9e0e5efb/qdev_wrappers/alazar_controllers/acquisition_parameters.py#L70) can just be done via `Parameter('name', instrument=instrument, get_cmd=None, set_cmd=False`, and funnily enough the `alternative` argument and `_alternative` attribute are not used at all, also in that module (but i might be missing stuff)

qdev_wrappers/alazar_controllers/acquisition_parameters.py
```diff
@@ -37,10 +37,9 @@ def set_raw(self, value):
...
         self._check_and_update_instr(value, param_name=self.name)
-        self._save_val(value)
 
     def get_raw(self):
```
astafan8: is anybody using [`aquisition_parametersold`](https://github.com/qdev-dk/qdev-wrappers/blob/3412bc01e13060ea88428a876477700d9e0e5efb/qdev_wrappers/alazar_controllers/acquisition_parametersold.py)? those also contain some `_save_val` calls. Perhaps that module can just be removed?

astafan8:
> Should we put a minimum qcodes version in the setup.py file?

In order to NOT forget Jens' comment above - we only have to do that if the proposed changes are backwards INcompatible. But per my evaluation, so far they are.

jenshnielsen:
replaced by #200

PR #200, opened 2020-08-20
Alazar update

FarBo:
- Alazar example notebook updated.
- Depreciated _save_val() is substituted with _cache_get() in ALazar controller files.
- Old example and controller files are deleted.


astafan8:
@FarBo be mindful about compatibility with old qcodes versions. At least, the minimal supported qcodes version has to be mentioned in the setup.py due to the usage of API that is in qcodes since ~0.12 or smth. Alternatively, one could look into preserving the compatibility with older qcodes versions, but that requires using API that is avaialbe in both new and older versions.

FarBo:
> @FarBo be mindful about compatibility with old qcodes versions. At least, the minimal supported qcodes version has to be mentioned in the setup.py due to the usage of API that is in qcodes since ~0.12 or smth. Alternatively, one could look into preserving the compatibility with older qcodes versions, but that requires using API that is avaialbe in both new and older versions.

@astafan8, Thanks Mikhail for the point. I will conduct a research and some tests to see which qdodes versions are compatible with these changes.

FarBo:
@astafan8 
Current qcodes version in setup.py is 0.1.9.
Parameter.cache.get which we used in this PR, first mentioned in the qcodes 0.9.0.  Should we consider to change the qcodes version to 0.9.0 in setup.py?

astafan8:
> Should we consider to change the qcodes version to 0.9.0 in setup.py?

yes. either this OR make the PR changes such that they still work with qcodes 0.1.9. I don't think we should do the latter, so indeed, just bump the minimal version to 0.9.0

qdev_wrappers/alazar_controllers/acquisition_parameters.py
```diff
@@ -21,9 +21,8 @@ class AcqVariablesParam(Parameter):
 
     def __init__(self, name, instrument, check_and_update_fn,
                  default_fn=None, initial_value=None):
-        super().__init__(name)
+        super().__init__(name, initial_cache_value=initial_value)
```
astafan8, line 24: it might be that initial_cache_value was introduced in version later than 0.9.0

FarBo: There is no specific mentioning of initial_cache_value in qcodes change logs. But, I found this:
initial_cache_value – Value to set the cache of the parameter to at the end of its initialization (this is equivalent to calling parameter.cache.set(initial_cache_value) after parameter initialization).
So I think having qcodes 0.9.0 as the minimum qcodes version should not be an issue regarding this initial_cache_value.

jenshnielsen: @FarBo Have a look here https://qcodes.github.io/Qcodes/changes/0.15.0.html#improved :)

astafan8: > There is no specific mentioning of initial_cache_value in qcodes change logs

unfortunately it's mentioned as initial_cache_val in 0.15.0 changelog https://github.com/QCoDeS/Qcodes/releases/tag/v0.15.0

@FarBo 

Another way to figure out when the change was introduces is to use "blame" (in github or in git itself) on the parameters.py file, and then the cool feature of github that shows tags that include a given commit, which results in this:

image

So initial_cache_value is there from qcodes 0.15.0. Hence if it is to be used, then the minimal version needs to be bumped to 0.15.0.

FarBo: I see. It is mentioned as initial_cache_val in the change log, and that was the reason I couldn't find it. Blame method is very cool, though. thanks Mikhail for the point.
I bump the version to 0.15.0 and make a new commit.

FarBo:
@astafan8 
Could you review the last commit? I am looking forward to have the changes to be merged in order to test it in local setups.

jenshnielsen (approved):

astafan8 (commented):
approve  :)

PR #201, opened 2020-08-24
Update test_doNd.py

FarBo:
This branch is the place to discuss about and/or change this test file.

for the first commit, I changed data.get_values to data.get_parameter_data in tests related to test_doNd_output_data(). I had to modify the related tests slightly because data.get_parameter_value returns the dataset in a dictionary format including numpy arrays for the values inside the dictionary.
Having these changes, pytest has passed all the tests locally.

### DIFF
--- a/qdev_wrappers/tests/dataset/test_doNd.py
+++ b/qdev_wrappers/tests/dataset/test_doNd.py
@@ -9,7 +9,7 @@ from qcodes import config, new_experiment, load_by_id
 from qcodes.utils import validators
 
 import pytest
-
+import numpy as np
 config.user.mainfolder = "output"  # set ouput folder for doNd's
 new_experiment("doNd-tests", sample_name="no sample")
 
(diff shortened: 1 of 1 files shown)


qdev_wrappers/tests/dataset/test_doNd.py
```diff
@@ -165,8 +164,8 @@ def test_do1d_output_data(_param, _param_set):
     data = load_by_id(exp[0])
 
     assert data.parameters == f'{_param_set.name},{_param.name}'
-    assert data.get_values(_param.name) == [[1]] * 5
-    assert data.get_values(_param_set.name) == [[0], [0.25], [0.5], [0.75], [1]]
+    assert np.allclose(data.get_parameter_data(_param.name)[_param.name][_param.name], np.array([1]*5))
```
jenshnielsen, line 167: Slightly cleaner way to write `np.array([1]*5))` with numpy would be `np.ones(5)`

jenshnielsen (approved):

qdev_wrappers/tests/dataset/test_doNd.py
FarBo: Good point, thanks. I will make the change accordingly and push it in a moment.

